PR #162, opened 2022-09-28
feature: Export group of packages (preview) 

harleylara:
Hi @martinet101

In the previous days I made sync of my fork and indeed many many things changed, which is amazing !!!

Well without further ado this is just the beginning of my contribution, as you can see at the moment only `winget` is supported as I am not an expert in `scoop` and I must continue to do some experiments.

Notes:
- With the new codebase and the renaming of some files I ran into some problems running `wingetUI` using the python interpreter, however I fixed all of them and it can run correctly.
- The checkbox looks different 😅, I just noticed it but I think it's something that can be easily fixed.

I'm curious to know your opinion: Don't you think that the `storeEngine` file is getting bigger and bigger and will get to a point where it becomes a bit tedious to manage?

Next steps: from my side I will keep reading the code, and maybe I will be helping to solve some issues, although I would also like to ask from your side what would you like some help on? docs? closing issues? something else?

With nothing more to add, best regards
- Harley Lara

### DIFF
--- a/wingetui/MainWindow.py
+++ b/wingetui/MainWindow.py
@@ -9,7 +9,7 @@ import win32gui
 import globals
 
 from storeEngine import *
-from tools import *
+from Tools import *
 
 class RootWindow(QMainWindow):
     callInMain = Signal(object)
--- a/wingetui/MainWindow.py
+++ b/wingetui/MainWindow.py
@@ -9,7 +9,7 @@ import win32gui
 import globals
 
 from storeEngine import *
-from Tools import *
+from tools import *
 
 class RootWindow(QMainWindow):
     callInMain = Signal(object)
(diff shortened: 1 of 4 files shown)


martinet101:
> Next steps: from my side I will keep reading the code, and maybe I will be helping to solve some issues, although I would also like to ask from your side what would you like some help on? docs? closing issues? something else?

Honestly, do whatever you want. If you feel like answering issues, fixing them or you just don't want to conttributte anymore it's all right! you have already helped a lot!

About the PR, i'll test the branch later and merge it into main.

martinet101:
> I'm curious to know your opinion: Don't you think that the `storeEngine` file is getting bigger and bigger and will get to a point where it becomes a bit tedious to manage?

Yes, i might end by dividing the UI sections and the package installers widgets

PR #229, opened 2022-10-11
install_scoop.cmd: Improve grammar

RavenMacDaddy:
image

On a more serious note, and regarding [previous](https://github.com/martinet101/WingetUI/issues/177) [discussions](https://github.com/martinet101/WingetUI/issues/101), some things to keep in mind:

* We want to keep the sentences as effective (i.e., short and clear) and at the same time as grammatically correct as possible – specifically because we're dealing with something being presented in a CLI environment.
* Why do we want to capitalize the first letter of Scoop? There are two things that are important to separate:

#

1. Scoop as a product name.

image

Notice how every instance is capitalized, as expected.

#

2. Working with Scoop, which is primarily through a CLI;

image

image

During operation, the typography is based on operating within a CLI environment, which in this case means only using lower case lettering.

---

So, we want to make the sentence more effective, and we want to describe Scoop in the form of a product name. 

Therefore, regarding the last part, we want to capitalize the first letter in this instance.

### DIFF
--- a/wingetui/resources/install_scoop.cmd
+++ b/wingetui/resources/install_scoop.cmd
@@ -1,5 +1,5 @@
 @echo off
-echo This script is going to install scoop and the dependencies required by WingetUI.
+echo This script will install Scoop and the dependencies required by WingetUI.
 pause
 powershell -NoProfile -Command "Set-ExecutionPolicy RemoteSigned -Scope CurrentUser"
 echo Installing scoop...


martinet101 (approved):

martinet101:
Hi @RavenMacDaddy, thanks for doing that. I have tons of stuff to do, and it is true that sometimes i leave behind some stuff like grammar or more case-specific features. 

Real thanks for fixing this little stuff that i hadn't even realised ;)

PR #267, opened 2022-10-15
Fix typo

ppvnf:
I believe this message is no longer displayed since WingetUI now uses --silent
Should it be removed?


martinet101:
No, this message is right. It gets triggered on installers that _ignore_ the --silent flag, and often when the user presses the `cancel` button in the program installer widget

martinet101 (approved):

PR #269, opened 2022-10-15
small text improvement

ppvnf:


martinet101:
Hi @ppvnf, regarding typos and corrections in the english language file, could you please correct them in the english file in Tolgee? It's because every time a string is modified on the source code it needs to be replaced on Tolgee. 

(p.d. you have now access to the english file)

Thanks for the amazing work you are doing!

PR #363, opened 2022-10-30
assign 'martinet101' like for bugs

ppvnf:
This difference between bugs and features wasn't intentional, right?


martinet101:
> This difference between bugs and features wasn't intentional, right?

nope

PR #365, opened 2022-10-30
Hungarian translation

gidano:
Based on the English language file downloaded from GitHub today


martinet101:
Hi, i can't merge that because it would break compatibility. 

Please check out the wiki (https://github.com/martinet101/WingetUI/wiki#translating-wingetui) in order to see how to translate WingetUI.

Thank you!

PR #491, opened 2022-12-13
Clarify that this option is specific to Scoop

ppvnf:


martinet101:
This option is not only for scoop, it also affects winget:

image

Is this actual feature the one that fixed issues with the data sources being held by another process

Issue #495, opened 2022-12-14
[BUG] Cannot install scoop git


panther7:
Probably need reload powershell/cmd enviroment (or restart app)

![2022-12-14 14 15 48](https://user-images.githubusercontent.com/3808250/207612906-bbd24293-0db2-41f1-965e-63e19a7019a4.png)

martinet101:
This issue should be now fixed!

panther7:
Maybe need restart app, or restart "detect" Scoop, if you install/uninstall Scoop.

Because in "About screen" is not refreshed information and "package list" is not refreshed.

martinet101:
Yes, WingetUI won't detect scoop until the next restart.
But the script shouldn't fail anymore when installing git.

I'm going to force-restart wingetui after the execution of the script

martinet101:
Done!

martinet101:
![image](https://user-images.githubusercontent.com/53119851/208185552-279b7588-291b-4402-af2d-ad6328c84d86.png)

PR #496, opened 2022-12-14
Fix install Scoop

panther7:
#495 

TODO:
* refresh detect Scoop
* refresh list of packages
* probably need restart app


wingetui/resources/install_scoop.ps1
```diff
@@ -0,0 +1,10 @@
+Write-Output "Installing scoop..."
+
+Invoke-Expression "& {$(Invoke-RestMethod get.scoop.sh)}"
```
marticliment: It is missing the flag to allow administrator installs, right?

marticliment: Done

martinet101 (approved):

PR #529, opened 2022-12-27
Draft: Improvements

panther7:

PR #531, opened 2022-12-27
Improvements

panther7:
ref #529


panther7:
Done, check it, if tis ok for you.

@marticliment

ppvnf:
Would it be possible to do this automatically to all translators (or at least the ones that have a Github account)?

panther7:
@marticliment Please try this fix [f20cb99](https://github.com/marticliment/WingetUI/pull/531/commits/f20cb99122580a68f677c87d429f9a9dc2080e97) for ISS.

panther7:
@ppvnf I can't check, if "name" is github accout. You can add "@" prefix for your name/nick and link will be created.

marticliment:
> @marticliment Please try this fix [f20cb99](https://github.com/marticliment/WingetUI/pull/531/commits/f20cb99122580a68f677c87d429f9a9dc2080e97) for ISS.

It seems to work for me. Thank you!

P.D. I can merge that, right?

panther7:
Yes, u can

ppvnf:
I've tested it and it works great. Thanks! Is there a simple way using Tolgee for @marticliment to add the other translators as contributors so they also have their Github profiles linked? Also, should I manually link my profile in the readme?

panther7:
Links from translators is not secured, I think, this is good solution ;-)

marticliment:
> I've tested it and it works great. Thanks! Is there a simple way using Tolgee for @marticliment to add the other translators as contributors so they also have their Github profiles linked? Also, should I manually link my profile in the readme?

Every contributor will need to add their username instead of their name

I'll add the ability to show also links in the readme (id you were to link that manually, it would get erased after the readme was generated again)

PR #561, opened 2022-12-30
Fix: Steam and MSStore aren't displayed in error message

ppvnf:
instead of showing only "Local PC", using f-string allows "Steam" and "MSStore" as well. I'm not sure, though, if it would be better to have a custom message for MSStore

Captura de tela 2022-12-30 113644
change to winget and scoop
Captura de tela 2022-12-30 113625


marticliment:
I would do  `_("Uninstallable [...] listed ad {0} can not be [...]").format(item.text(4))`, so the string `"Uninstallable [...] listed ad {0} can not be [...]"` can be easily added to tolgee

ppvnf:
Done

marticliment (approved):

PR #565, opened 2022-12-31
identify ubisoft connect

ppvnf:
**Missing icons, because I am not sure which one to use:**

uplay_white.png
uplay_black.png

Maybe use a generic spiral icon like 

uplay_black
uplay_white


marticliment:
icons8_ubisoft_512px

What about something like this? (it's their logo)

ppvnf:
I don't have any problem with using it, however couldn't there be a copyright issue?

marticliment:
No, it is an imitation of their logo made by icons8.com, and they distribute this icon as an usable icon under their terms. 
I don't see any issue

Issue #625, opened 2023-01-18
[FEATURE REQUEST] Fail dialog after installation which need restart


panther7:
After update LibreOffice

![2023-01-18 10 56 08](https://user-images.githubusercontent.com/3808250/213148750-87a3bfd6-f012-49af-b250-b4a584a17c61.png)

marticliment:
Fixed in your PR

PR #626, opened 2023-01-18
Handle need restart after installation for winget

panther7:
#625

Added some new texts, but question is, are we really need dialog?

@marticliment


marticliment:
I think this dialogs might be a little bit intrusive, especially when bulk-updating packages

panther7:
> I think this dialogs might be a little bit intrusive, especially when bulk-updating packages

I agree, but now, show "error" dialog.

Should I just rewrite it in a text message like "installed successfully"?

marticliment:
The current behaviour is:
1. If failed, show a message.
2. If succeeded, throw a notification.

What I would do is change the text of that "{0} installation succeeded" for "A restart is required to finish {0} installation", as well as addding a visual indicator in the ui telling the user to restart its PC.

If you don't understand how the error message class works, (it is a little bit messy), just let me know and I'll implement this feature

panther7:
Done, new strings:
- {action} was successfully!
- Restart your PC to finish installation

marticliment:
image
image

I have added new restart icons, so it looks visually different

marticliment (approved):

PR #631, opened 2023-01-22
Unified translation strings for package managers

panther7:
Unified translation string for pacakge managers, like:
- Locating {pm}...
- {pm} found: {state}
- {pm} preferences
- {pm} package manager specific preferences
- Enable {pm}

{pm} (package manager) = Winget/Scoop/Chocolatey ...maybe is good write as notice to Toolge
{state} = True/False

More untranslated strings:
- Chocolatey packages are being loaded. Since this is the first time, it might take a while, and they will show here once loaded.
- Blacklist packages
- Show Scoop packages in lowercase
- Local PC

### DIFF
--- a/wingetui/__init__.py
+++ b/wingetui/__init__.py
@@ -178,13 +178,13 @@ try:
 
         def detectWinget(self):
             try:
-                self.callInMain.emit(lambda: self.loadingText.setText(_("Locating Winget...")))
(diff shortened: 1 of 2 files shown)


marticliment (approved):

marticliment:
I have added the missing strings!

panther7:
> I have added the missing strings!

`Locating {pm}` ->`Locating {pm}...`

PR #904, opened 2023-04-11
Draft: Check translations

panther7:
I working for check translations... actully export is (too much nontraslated strings :-/) :

```json
{
  "not_used": [
    "{0} {0} {0} Contributors, please add your names/usernames separated by comas (for credit purposes)",
    "Add",
    "Available updates",
    "Enable Scoop",
    "Run as administrator",
    "Uninstall selected package",
    "WingetUI version {0}"
  ],
  "not_translated": [
    "{0} aborted",
    "{0} packages were found",
    "{0} package was found",
    "Cache was reset successfully!",
    "Do you really want to uninstall {0} packages?",
    "Instance {0} responded, quitting...",
    "It looks like you ran WingetUI as administrator, which is not recommended. You can still use the program, but we highly recommend not running WingetUI with administrator privileges. Click on \"{showDetails}\" to see why.",
    "The checksum of the installer does not coincide with the expected value, and the authenticity of the installer can't be verified. If you trust the publisher, {0} the package again skipping the hash check.",
    "The installer has an invalid checksum",
    "We could not {action} {package}. Please try again later. Click on \"{showDetails}\" to get the logs from the uninstaller.",
    "We could not load detailed information about this package, because it was not found in any of your package sources"
  ]
}
```

@marticliment

### DIFF
--- a/.github/workflows/update-tolgee.yml
+++ b/.github/workflows/update-tolgee.yml
@@ -15,6 +15,9 @@ jobs:
       - name: Download Tolgee translations
         run: |
           python scripts/download_translations.py --autocommit
+      - name: Check translations
+        run: |
+          python scripts/export_translation_strings.py
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v5
         with:
@@ -25,3 +28,4 @@ jobs:
           author: "Tolgee Bot <tolgee-bot@users.noreply.github.com>"
           commit-message: "Update translations from Tolgee"
           branch: pull-request/update-translation
+#          body-path: 
--- a/scripts/create_translations.py
+++ b/scripts/create_translations.py
@@ -1,0 +1,71 @@
+import argparse
+import tolgee_requests
+import translation_utils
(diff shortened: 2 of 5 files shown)


marticliment:
Those that include \"{showDetails}\", they can be found without the backslashes in Tolgee.

panther7:
> Those that include "{showDetails}", they can be found without the backslashes in Tolgee.

Sorry, but my script works fine, "all" `not_translated` are relevant, you can check manualy.

flatron4eg:
> P.D. There's actually an utility on /scripts that will check the code for unused strings.

This script is trying to find just the strings without the quotes and braces, so it can find line "Available updates" in code cause we have line "Available updates: {0}, not finished yet..." (so it's just a substring of actual line), and first one is not used actually (so as all the others that @panther7 mentioned), his script is good. I was working on fixes for your script but then found this PR =)

panther7:
Created new scripts via #926.

PR #934, opened 2023-04-19
Fixed closeEvent dialog, when ignore user choise if installation was …

panther7:


wingetui/mainWindow.py
marticliment, line 1: Definitely a needed restructuration, I still don't understand how my code turned up so messy...

Thanks!

marticliment (approved):

PR #970, opened 2023-04-29
Add suggestion to scoop info menu

ppvnf:
I am not sure I used the most efficient method

Closes https://github.com/marticliment/WingetUI/issues/602


marticliment:
Chocolatey also lists dependiencies (and winget will do in the next version), and this new entry could be a good place to list them. What do you think?

ppvnf:
I didnt know about that. I agree

ppvnf:
I guess https://github.com/marticliment/WingetUI/issues/834 and https://github.com/marticliment/WingetUI/issues/602 are related then, since dependencies in Chocolatey needs manifest parsing. Also, I think that dependencies is a better name than suggestion

PR #1000, opened 2023-05-03
Bunch of f-string, with statements / os.path usage changes

Mikey1993:
- [X] **I have read the [contributing guidelines](https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#coding) and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**

-----
I have made changes to the python scripts:
- Use f-strings where is needed, and removed where it's not
- Converted all the os.open() statements to use the with statement for a better code etiquette 
- Made the use of `os.path.normpath()` and `os.path.join()` where is needed for better cross platform comparability and robustness

-----


Mikey1993:
@marticliment I think these changes are pretty trivial, do you see any issues with merging them?

marticliment (approved):

panther7:
OMG, please try before merge... @marticliment 

```
Traceback (most recent call last):
  File "C:\Users\filip\git\WingetUI\scripts\get_contributors.py", line 38, in <module>
    contributors_filepath = os.path.normapth(os.path.join(root_dir, "wingetui/data/contributors.py"))
                            ^^^^^^^^^^^^^^^^
AttributeError: module 'ntpath' has no attribute 'normapth'. Did you mean: 'normpath'?
Traceback (most recent call last):
  File "C:\Users\filip\git\WingetUI\scripts\download_translations.py", line 32, in <module>
    os.chdir(os.path.normapth(os.path.join(root_dir, "wingetui/lang")))
             ^^^^^^^^^^^^^^^^
AttributeError: module 'ntpath' has no attribute 'normapth'. Did you mean: 'normpath'?
```

Btw, whats was wrong before? @Mikey1993

Mikey1993:
@marticliment ,Sorry for the mess..
- It's still better to use os.path.join() than hard coding it
- Same goes for the `with` for opening files.
- Regarding the os.path.normpath(), I get what you say @panther7, but this is a good practice to use nevertheless. Also, you practically should be able to run this code on a non Windows machine as well (the python scripts), so this does fix an issue.

PR #1353, opened 2023-09-12
Add an interface to filter packages by source

### LINKED ISSUE #584: [FEATURE REQUEST] Button to filter source
## Is your feature request related to a problem? Please describe.

I think it would be useful to have a button that allows only a certain source to be shown (Scoop, Winget, Local PC, etc.)

## Describe the solution you'd like

Examples:

1. In the discovery section, selecting the option to filter Winget (possibly using a combo box) would hide all pkgs from Scoop (instead of having to go to settings and disable Scoop). EDIT: maybe selecting the scoop bucket would be useful as well.

2. For the update section I thought about having an option to update all Scoop / all Winget packages separately, but I personally just update everything regardless, so I don't think I would use it.

3. In the uninstall section it would be useful for seeing all packages from MSStore or from a specific bucket.

The button could be similar to https://github.com/marticliment/WingetUI/issues/223 and https://github.com/marticliment/WingetUI/issues/339

Edit:
If you think it is useful/easy to implement (I'm not sure actually) this button could also enable/disable case-sensitive search
Maybe you could also include instant search and show unknow version (update section) inside of it as well to optimize space utilization

Icon suggestion (a simple arrow pointing down, not necessarily this one):

<details>

long-arrow-down

</details>

### LINKED ISSUE #822: [FEATURE REQUEST] Show a breakdown of how many apps are shown in the table from each source in the UI
## Before submitting the issue

- [X] **I have checked that the requested feature is not a duplicate or, if it is, it needs to be reopened**
- [X] **I have checked that I am running the latest version of WingetUI**
- [X] **I have checked the [FAQ](https://github.com/martinet101/WingetUI#faq) section**
- [X] **I have read the [contributing guidelines](https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#feature-requests) and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**

## Describe the solution you'd like

I suggest adding a breakdown of the packages by source like so:
1. In the "Discover Packages" and "Software Updates" tab breakdown the packages to the sources that are currently loaded from the table (Winget/Chocolatey/Scoop) something like that:
image

2. In the "Installed packages" do the breakdown the same for all the sources (even for the Local PC/Steam/Microsoft Store ones for better visibility).
image

marticliment:
Preview of the feature:
image

fix #822 
fix #584


neoOpus:
Thank you Marti :)

PR #1588, opened 2023-11-25
Add custom WebView2 wrapper to show help articles

marticliment:
Add a C#.NET WebView2 wrapper to show help articles instead on showing them on the browser

TODO: 

- [x] getUrl method needs fixing


raghavdhingra24:
Hi Marti,
I wanted to ask why this approach is being used over Qt's [QWebEngineView](https://doc.qt.io/qt-6/qwebengineview.html) as it might get harder to maintain in future (Just a speculation)

marticliment:
This choice is done because:
1. QtWebEngineView is bundled with the app and occupie 150+ mb of disk space. It requires updating the app in order to update the engine. Instead, WebView2 is bundled with Windows 10/11, so it does not require to be updated regularly from the app and it only occupies ~6mb.
2. QtWebEngineView is laggy and is not optimized. WebView2 uses hardware acceleration to smoothen navigation so users can enjoy a faster, more pleasant experience.
4. This is a test that will serve to see to which extent embedding C#.NET windows on a Qt window is viable and if it works flawlessly. If the results are succesful on this non-critical feature, more parts of the interface may get migrated to WinUI3 windows, in order to provide a better interface. 

In theory, maintaining should not be harder as long as Qt does not remove the feature to embed windows given their handle. Moreover, the ability to work with two independent environments (Qt and WebView) means the application is not limited to what Qt can do. In fact, the share interface already uses this method, and as far as I know it has been working without any issue.

PR #1813, opened 2024-02-24
Update dependency

Malus-risus:


marticliment:
What does renovate.json do?

Malus-risus:
> What does renovate.json do?



> What does renovate.json do?

Renovate.json is used for dependency updates and is now deprecated.

marticliment:
And if it is deprecated shouldn't it be removed?

Malus-risus:
> And if it is deprecated shouldn't it be removed?

okay

marticliment (approved):

PR #1823, opened 2024-02-26
Refined Form1 Loading & Visibility Mechanics in WinFormsWebView

Malus-risus:
### Added
- SemaphoreSlim `_loadSemaphore` to manage concurrent access to the `loaded` flag.

### Changed
- Replaced `getHWND()` method with a read-only property `HWND`.
- Renamed `uncoverWindow()` to `UncoverWindow()` to follow C# naming conventions.
- Encapsulated `loaded` field with a private modifier and renamed it to `_loaded`.
- Introduced `WaitForLoadAsync()` private method to wait for the WebView to load, reducing code duplication.
- Updated navigation methods (`NavigateTo`, `NavigateToString`, `Stop`, `Reload`) to use `WaitForLoadAsync()`.
- Converted `getUrl()` method to a read-only property `Url`, reducing complexity.
- Removed redundant `loaded = true;` after starting navigation to `about:blank`.
- Simplified exception handling in `LoadWebView()` method to use `when` clause for specific exceptions.
- Added `NavigationCompleted` event handling to set `_loaded` flag to true.
- Streamlined window visibility changes by using consistent C# methods `Show()`, `Hide()`, and `Opacity` property.

### Fixed
- Ensured `LoadWebView()` only sets `_loaded` flag within a fully awaited and locked context using a semaphore.
- Addressed potential concurrency issues by introducing a semaphore for loading-related operations.

### DIFF
--- a/components/webview2-component/c#/Form1.cs
+++ b/components/webview2-component/c#/Form1.cs
@@ -1,14 +1,6 @@
-﻿using Microsoft.Web.WebView2.Core;
+using Microsoft.Web.WebView2.Core;
 using System;
-using System.Collections.Generic;
-using System.ComponentModel;
-using System.Data;
-using System.Diagnostics;
-using System.Drawing;
 using System.IO;
-using System.Linq;
-using System.Text;
-using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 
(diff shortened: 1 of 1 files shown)


marticliment:
Hello @Malus-risus,

the webview component is also deprecated. 

All the current code can be found on pr #1724 or branch winui-widgets

PR #1876, opened 2024-03-08
3.0.0 Beta 1 - Fix System.IO.DirectoryNotFoundException when trying to save package installation options

dave-sc:
While testing #1875 I noticed that I am unable to save any changes made to the installation options of a chocolatey package, as my `%USERPROFILE%\.wingetui` directory does not have a subfolder the app is trying to write to. This change ensures that the parent directory actually exists.

I'd also suggest to use the `FileInfo` class instead of the static methods, as I stumbled upon this pattern:
``` csharp
if (!Directory.Exists(Path.GetDirectoryName(__ignored_updates_database_file)))
    Directory.CreateDirectory(Path.GetDirectoryName(__ignored_updates_database_file));
if (!File.Exists(__ignored_updates_database_file))
    File.WriteAllText(__ignored_updates_database_file, "{}");
```
which would nicely clean up to:
``` csharp
var ignoredUpdatesFile = new FileInfo(__ignored_updates_database_file);
if (ignoredUpdatesFile.Directory?.Exists == false)
    ignoredUpdatesFile.Directory.Create();
if (!ignoredUpdatesFile.Exists)
    File.WriteAllText(ignoredUpdatesFile.FullName, "{}");
```
Less juggling of file paths and no compiler warning, as `Path.GetDirectoryName` *technically* can return null.

- [X] **I have read the [contributing guidelines](https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#coding) and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**

-----

-----


marticliment:
Looks good to me.
I will merge the PR once I test it

marticliment (approved):

PR #1881, opened 2024-03-09
Fixed a typo

DariusHutchison:
- [x] **I have read the [contributing guidelines](https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#coding) and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**

### DO NOT publish garbage PRs to farm Crypto AirDrops. Any user suspected of this actions will get banned. Submitting broken code wastes the time of the contributors, who have to spend their free time reviewing, fixing and testing code that does not even compile, does break other functionality or does not introduce any changes at all.

-----

-----

Closes #XXXX

Relates to #XXXX


marticliment:
This issue is fixed by setting the value of the correct key on the language file, and not by changing the key itself

PR #1962, opened 2024-03-15
SoftwarePage width

Schtenk:
Changes:
Changed rootgrid column widths on SoftwarePages to make the pages populate more of the available space on wider window sizes.

Notes
Not sure about your intent for the look of these pages are but this should change them to keep the same spacing to the edges of the window unless made thinner than the minimum sizes.

Before:
image

After:
image

- [ ] **I have read the [contributing guidelines](https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#coding) and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**
- [ ] **Have you checked that there aren't other open [pull requests](https://github.com/marticliment/WingetUI/pulls) for the same changes?**
- [ ] **Have you tested that the committed code can be executed without errors?**

### DO NOT publish garbage PRs to farm Crypto AirDrops. Any user suspected of this action will get banned. Submitting broken code wastes the time of the contributors, who have to spend their free time reviewing, fixing and testing code that does not even compile, breaks other features or does not introduce any changes.

-----

-----

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/src/wingetui/Interface/SoftwarePages/DiscoverPackages.xaml
+++ b/src/wingetui/Interface/SoftwarePages/DiscoverPackages.xaml
@@ -583,13 +583,13 @@
             <RowDefinition Height="*"/>
         </Grid.RowDefinitions>
         <Grid.ColumnDefinitions>
-            <ColumnDefinition Width="*"/>
-            <ColumnDefinition Width="1000000*" MaxWidth="1350"/>
-            <ColumnDefinition Width="*"/>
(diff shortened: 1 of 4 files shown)


marticliment:
IMO having the colums being too wide can make it more diffucult to read packages, but obiously the limit I had set was too small. I will increase it to 2000

marticliment (approved):

PR #1970, opened 2024-03-16
Dotnet runtime bootstrapper

marticliment:
Avoid having to bundle .NET and VCRedist installers with WingetUI installer.

This code library requires further evalutation before being released to the general public


marticliment:
Will be replaced by this:

https://github.com/DomGries/InnoDependencyInstaller

marticliment:
Reduce installer size from 160MB to 17MB (deps are downloaded on-the-go if needed)

PR #2166, opened 2024-04-24
Refactoring for agility and testability

Atrejoe:
As agreed by mail, this should aid in making the code more testable by moving core logic away from UI.
These are only the first steps.

- [x] **I have read the [contributing guidelines](https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#coding) and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**
Any user suspected of farming GitHub activity with crypto purposes will get banned. Submitting broken code wastes the time of the contributors, who have to spend their free time reviewing, fixing, and testing code that does not even compile, breaks other features, or does not introduce any useful changes. Thank you for your understanding

-----

A lot of logic was interdependent, I have tried to split this into more single-purpose, injectable classes.

-----

Relates to #2161


Atrejoe:
I'm looking at your branch as we speak.
I expect them not to be compatible, but I could give it a go.



On Wednesday 24 April 2024 12:45 CEST, Martí Climent ***@***.***> wrote:
 
 
Hi @Atrejoe, I have just started doing that on #2161, i think I did not explain myself well on the email😬.
Do you think you can redirect the PR to the split-into-separate-projects branch so I can take advantage of your work? Right now I have isolated the core/non-ui and non-packageengine part of the software, and I have some tests set up and some more under development.

marticliment:
Don't worry about compatibility, I will fix conflicts and figure everything out

Atrejoe:
Hi Martin,

I did two rebasing sessions, in which many of the conflict I have were you and me addressing the same issue in a different way.
Both rebasing session failed due to git (got into a detachd head state, where git forgot it was rebasing).

In general the difficult parts, splitting the AppTools class into multiple classes addressing different concerns (moving logging, configuration and generic utility functions to new classes), you did more or less the same thing. For simplicity's sake I chose your option, leaving little of my initial refactoring.
My method made the first steps towards dependency injection, making the solution more composable, but that can also be done at a later stage. You have slit into multiple libs, that was the hard parts.

You could argue that you went a little overboard, and now there are to many libs, but that is easier to solve.

Which parts of my PR did you consider worthy of salvaging?
Maybe my efforts are better concentrated on converting your current unit test to XUnit and introduing testcases and general refactoring.

Regards,

Robert






On Wednesday 24 April 2024 12:53 CEST, Martí Climent ***@***.***> wrote:
 
 
Don't worry about compatibility, I will fix conflicts and figure everything out

marticliment:
Well, I haven't started yet separating the PackageManager stuff from the UI stuff yet, and I see you already have done a little bit on that side, so that is the thing I guess would make the most sense to integrate on the new branch. 
However, this may be difficult due to the already done progress. If you prefer to focus on XUnit (I have looked at it, and it looks promising) and on modifying already existing tests and/or introducing new/missing ones, I think it might be better and easier for both of us; and I will take care of the rest of the splitting into smaller projects.

> You could argue that you went a little overboard, and now there are to many libs, but that is easier to solve.

It is my first time not using a huge project, so I still need to get used to it, and I think you are right, but I preferred to oversplit and then be able to merge projects afterwards rather than undersplitting and having to deal with more splitting again later.

PR #2524, opened 2024-07-20
Add base (non-uft8 compatible yet) redirector

marticliment:
Create a simple c++ WingetUI.exe executable that launches UniGetUI.exe, so process name is shown as UniGetUI


marticliment:
Not needed anymore

PR #2662, opened 2024-08-24
Add Cargo package manager

wilt00:
-----
This PR adds basic support for Cargo, the Rust package manager.

Remaining blockers keeping this PR as a draft:

- [x] Currently, the process for installing and upgrading packages seems to hang without terminating, and the fact that logs aren't refreshed in the app until the end of execution is making this hard for me to debug. (The Live Output doesn't seem to be.) At one point I had this working, and I'm not sure what I changed to break it, so I'm posting this PR as-is in the hopes that you can spot what I'm doing wrong.
- [x] Missing package manager icon asset
- [ ] Documentation
-----


marticliment:
Great! if you need help or guidance, please let me know!

marticliment:
And this is why that happens

marticliment:
Operations are working now. 

Do you think it would be possible to not show libraries on the Discover packages page? I mean, show only binary-containing packages?

wilt00:
Thanks! I've added functionality to check each search result against the registry api to see if it has any binaries listed, which isn't perfect since some libraries include test binaries, but should handle most cases. Unfortunately it is quite slow, to comply with the [one request per second](https://crates.io/data-access) rate limit.

marticliment:
And do you think it would be possible to fetch this data directly from the https://github.com/rust-lang/crates.io-index ? In my understanding, since you can obtain the package's manifest, you should be able to parse it, right?

wilt00:
Unfortunately the index files at that link don't include the `bin_names` field that indicates if a package has binaries. Another thing that would work, though, is getting that info from the daily DB dump at https://static.crates.io/db-dump.zip . That would either require users to download and parse a ~700MB file, or a small backend service to do the same thing and expose a lighter API. Possibly a github action?

marticliment:
And if we parse the crates.io website? I have found that on binary packages, the install command is `cargo install`, while on libraries it is `cargo add`:

See here:
 - https://crates.io/crates/rls (bin)
 - https://crates.io/crates/exon-vcf/0.32.0 (lib, i suppose)

marticliment:
The page html does contain the info, so this should be doable:

image

marticliment (approved):

PR #2714, opened 2024-09-08
Winget COM API - Allow lower trust registration

Schtenk:
- [x] **I have read the [contributing guidelines] (https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#coding), and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**.
-----

Fixes winget class registration failing when running as admin by allowing lower trust registration.

Adds a toggle for allowing lower trust registration in winget section of UnigetUI settings.

-----

Closes #XXXX

Relates to #XXXX


Schtenk:
From my limited understanding of the windows registry, light research and personal testing it seems that winget COM API only registers as a lower than admin trust (or current user, i.e. not local machine wide), because of this allowing lower trust registration is needed when running as admin.

marticliment:
Do you think enabling this setting by default is a security concern? I couldn't find any docs on microsoft learn, I'll continue my research though

Schtenk:
I don't feel I can confidently say since I don't know all the ins and outs of the windows registry and user privileges in windows.
The documentation and mentions about the flag seems very scarce.

The first search result in google and the only usage I could find was in [ComUtils.cs](https://github.com/microsoft/winget-cli/blob/master/src/Microsoft.Management.Deployment.Projection/Utils/ComUtils.cs) from the WingetCLI github repo.
Which is then used through [LocalServerInstanceInitializer.cs](https://github.com/microsoft/winget-cli/blob/master/src/Microsoft.Management.Deployment.Projection/Initializers/LocalServerInstanceInitializer.cs)
in the test in [InstanceInitializersSource.cs](https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerCLIE2ETests/Interop/InstanceInitializersSource.cs) all from said repo.

marticliment (approved):

PR #2786, opened 2024-09-27
Fixed pre-release workflow

vikingnope:
-----

I found that the GitHub action did not run with your latest pre-release. After some more digging I found the the prereleased type is a bit misleading. Following this PR change it should actually work now.

image

-----


marticliment (approved):

marticliment:
Thanks for catching it, I was wondering why it wasn't running the workflow!

Issue #2883, opened 2024-10-23
[BUG] Subtitles and Checkboxes almost unreadable in dark mode [hot theme change only]


ghost:
### Please confirm these before moving forward

- [X] I have searched for my issue and have not found a work-in-progress/duplicate/resolved issue.
- [X] I have tested that this issue has not been fixed in the latest [(beta or stable) release](https://github.com/marticliment/WingetUI/releases/).
- [X] I have checked the [FAQ](https://github.com/marticliment/WingetUI#frequently-asked-questions) section for solutions.
- [X] This issue is about a bug (if it is not, please use the [correct template](https://github.com/marticliment/WingetUI/issues/new/choose)).

### UniGetUI Version

3.1.2

### Windows version, edition, and architecture

Windows 11 Pro 23H2 x64

### Describe your issue

Subtitles and checkboxes are almost unreadable in darkmode

### UniGetUI Log

```text
[23/10/2024 20:53:15]    __  __      _ ______     __  __  ______
                        / / / /___  (_) ____/__  / /_/ / / /  _/
                       / / / / __ \/ / / __/ _ \/ __/ / / // /
                      / /_/ / / / / / /_/ /  __/ /_/ /_/ // /
                      \____/_/ /_/_/\____/\___/\__/\____/___/
                          Welcome to UniGetUI Version 3.1.2
[23/10/2024 20:53:15]
[23/10/2024 20:53:15] Version Code:  3,12
[23/10/2024 20:53:15] Encoding Code Page set to 850
[23/10/2024 20:53:17] Using bundled GSudo at C:\Users\rober\AppData\Local\Programs\UniGetUI\Assets\Utilities\gsudo.exe
[23/10/2024 20:53:20] Tried to access a non-existing cached language file! file=C:\Users\rober\AppData\Local\UniGetUI\CachedLanguageFiles\lang_it.json
[23/10/2024 20:53:20] Loaded language locale: it
[23/10/2024 20:53:23] Randomly-generated background API auth token for the current session: vrijbqyqnmwb0jv5ql5s2uekdik49anrx8jqod0du04ck8epvopa449uqt5umvsp
[23/10/2024 20:53:23] Api running on http://localhost:7058
[23/10/2024 20:53:23] Lang files were updated successfully from GitHub
[23/10/2024 20:53:23] UniGetUI is up to date
[23/10/2024 20:53:24] The restored geometry did not have exactly 5 items (found length was 1)
[...]

PR #2925, opened 2024-11-03
Vcpkg Package Manager

### LINKED ISSUE #2146: [FEATURE REQUEST] VCPKG support?

- [X] This proposal is a completely new feature. If you want to suggest an improvement or an enhancement, please use [this template](https://github.com/marticliment/WingetUI/issues/new?assignees=marticliment&labels=enhancement&projects=&template=enhancement-improvement.yml&title=%5BENHANCEMENT%5D+%28Enter+your+description+here%29)

### Describe the new feature

There is support for pip, npm. It would be cool to have vcpkg support :)

### Describe how this new feature could help users.

vcpkg is one of the most popular package managers for C++

mrixner:
-----

# TODO before merging (for @marticliment)
 -  [x] Add the VCPkg icon to symbol icon table
 -  [x] Add the setting to specify the default triplet.

-----

This pull request is for support for the vcpkg package manager.

I'm sorry if the code is not good, this is my first time ever working with C#. I tried to write code in the style of all the other package managers and not touch more than I needed to.

## List of Added Translations:
- vcpkg ManagerProperties Description (A popular C/C++ library manager. Full of C/C++ libraries and other C/C++-related utilities&lt;br&gt;Contains: &lt;b&gt;C/C++ libraries and related utilities&lt;/b&gt;)
- InvalidOperationException value (Cannot update vcpkg port files as requested: git was not installed or the VCPKG_ROOT environment variable / the custom vcpkg root setting were not set)
- Automatic Portfile Update Setting Title (Update vcpkg's Git portfiles automatically (requires Git installed))
- Default vcpkg Triplet Setting Title (Default vcpkg triplet)
- Custom vcpkg Root Setting Title (Custom vcpkg root)

## Manager Capabilities (for the table on the home page):

**Install as administrator**: yes (for vcpkg roots in an admin-locked directory)
**Skip integrity checks**: no (I don't think vcpkg has this capability, or if it has integrity checks to skip)
**Interactive installation**: no (it's being built from the command-line; there's no interactivity to show)
**Install Older Versions**: no (vcpkg doesn't support this on a global scale, only for projects with [vcpkg.json](https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json) control files)
**Install a PreRelease Version**: no (vcpkg doesn't have pre-release versions)
**Install a Custom Architecture**: not directly supported but can be easily achieved (via using a different triplet / source)
**Install on a Custom Scope**: no (everything just gets installed to the vcpkg directory)
**Custom Install Location**: no / sort of not directly supported but can be easily achieved? (everything just gets installed to the vcpkg directory; you can change which vcpkg directory via the "Custom vcpkg root" setting)
**Custom Package Sources**: not directly supported but can be easily achieved (it will automatically detect custom triplet / source CMake scripts dropped into `%VCPKG_ROOT%\triplets` or `%VCPKG_ROOT%\triplets\community`)

## Other Potentially Unresolved Things

- I have added a colored icon, but as [src/UniGetUI.PackageEngine.Managers.Vcpkg/Vcpkg.cs:65](https://github.com/mrixner/UniGetUI/tree/vcpkg-package-manager/src/UniGetUI.PackageEngine.Managers.Vcpkg/Vcpkg.cs) states, discussion #2826 brought me to believe that an outlined icon must be added by the maintainers, so if you care enough that may be one thing that needs to happen.
- I would recommend probably squash-merging this if you weren't going to already, as I pretty much just edited the file for package manager-specific settings and the files for the package manager. Unfortunately, I accidentally forgot to commit some changes for some commits, so not all commits compile as requested in the Contributing Guidelines.
- As mentioned in #2146: The `DefaultVcpkgTriplet` setting isn't being set properly, but I was told to leave the UI to Marti, and so that's what I'm doing.

-----

Closes #2146


marticliment:
Hey, I have been looking at the code, and I have found this:
image

There is an overridable method on the PackageManager class (`public void RefreshPackageIndexes()`) that will be called by UniGetUI when needed to update the sources. 
I will move the code there, but I reckon if you then think the `Settings.Get("UpdateVcpkgGitPorts")` is still needed or not

marticliment:
Also, do you think it would make sense to list `git` as a dependency for vcpkg, so if the user doesn't have git installed UniGetUI prompts them to install it?

marticliment:
VcpkgSourceProvider.cs
image


What is the idea here?
GetAddSourceParameters and GetRemoveSourceParameters are supposed to return the command-line paramaters to add or remove a source, which I understand does not make sense when talking about triplets. Therefore, `_getAddSourceOperationVeredict` and `_getRemoveSourceOperationVeredict` would be never called, right?

marticliment:
Please try the changes I have pushed to confirm they work for you.
When we resolve the questions/comments mentioned above, I will merge the PR

mrixner:
Sorry for the late response (and for making you read all that code...).

> There is an overridable method on the PackageManager class (`public void RefreshPackageIndexes()`) that will be called by UniGetUI when needed to update the sources.

Sorry, I didn't realize that.

> I reckon if you then think the `Settings.Get("UpdateVcpkgGitPorts")` is still needed or not

I'm not sure. I guess it depends on how much control you want to give the user; I assumed not everyone would want the vcpkg repository getting automatically updated, but I suppose you're right; it doesn't matter.

I see you changed the setting to `DisableUpdateVcpkgGitPorts`; I think that makes more sense. I feel like you should be able to say "no, don't automatically update my git repository", in case maybe you're working in it or have it at a certain commit for some reason? But updates should be enabled by default. I think the current implementation with `DisableUpdateVcpkgGitPorts` is fine, unless you don't want the setting, as it's not strictly necessary to the operation of vcpkg.

> Also, do you think it would make sense to list `git` as a dependency for vcpkg, so if the user doesn't have git installed UniGetUI prompts them to install it?

I don't think so. Aside from the fact that people using vcpkg probably already have git installed, given that you need to have cloned the git repository in order for vcpkg to work properly git will probably be installed if they have vcpkg. Although - it won't hurt anything and is probably better for clarity, so maybe.

> What is the idea here?
> GetAddSourceParameters and GetRemoveSourceParameters are supposed to return the command-line paramaters to add or remove a source, which I understand does not make sense when talking about triplets. Therefore, _getAddSourceOperationVeredict and _getRemoveSourceOperationVeredict would be never called, right?

I don't know if they're called; I'm not sure how source management works. I assumed they would be called anyway when a source was added in the settings, and so the code there will return whether or not a given source is valid (recognized by vcpkg).

I tried removing the custom sources ability, since the code should autodetect any added triplets, but it removed the triplet sorting in the Available Updates section which I thought should be there, but you're right; those four functions don't _really_ serve a purpose.

I see you removed its ability for custom sources. This doesn't change any of the actual functionality; again, the only reason _I_ didn't do this was for the by-triplet selection in the available updates list.

> Please try the changes I have pushed to confirm they work for you.
> When we resolve the questions/comments mentioned above, I will merge the PR

I'll look at the changes and see if it works. I'll also change some of the other things I found.

mrixner:
OK, looking through the UI, the settings UIs look really great, and everything major works as intended. However, there's a few things I noticed.

- The default triplet dropdown shows a restart required banner when it doesn't need to (fixed in 2e5a166)
- I was hoping the first time you viewed the default triplet dropdown the value would be the default triplet as determined elsewhere, however this is a tiny nitpick and I understand if it's not possible.
- In the "discover" list of searched packages, if you have a package installed for a different triplet than the one you're searching for it still shows the little "installed" icon. I'm not sure what it is using to compare that, but I assumed that given that the package IDs are different, that would be different as well?
- As mentioned in the last comment, due to removal of custom sources, you can no longer filter the vcpkg packages by their triplet. This is also trivial and possibly intended / doesn't matter, but if there's an easy way to add that I think that could be helpful.

I'll look through all the diffs to see if there's anything else I missed.

mrixner:
I removed the unnecessary restart required banner, fixed a typo, and made `Vcpkg.GetDefaultTriplet` public. I did not remove any of the TODOs dealing with `DisableUpdateVcpkgGitPorts`, in case it's not something you want to keep.

I noticed this in the commit log; is there an implementation to think about? (it's been shifted to SettingPage`.xaml.cs:441`)
image

marticliment:
> * As mentioned in the last comment, due to removal of custom sources, you can no longer filter the vcpkg packages by their triplet. This is also trivial and possibly intended / doesn't matter, but if there's an easy way to add that I think that could be helpful.

Oh, this actually makes sense. 
What I will do is I will add source compatibility again, but I will manually remove the Source Manager UI from the settings window.



> * I was hoping the first time you viewed the default triplet dropdown the value would be the default triplet as determined elsewhere, however this is a tiny nitpick and I understand if it's not possible.

This can be _kinda_ fixed. On the LoadManager() method, you can check if the setting is empty, and if it is not set, it can be set to a default value. I don't know how to get the system default triplet, but if you know how to and you want to fix this it'd be great, 


> * In the "discover" list of searched packages, if you have a package installed for a different triplet than the one you're searching for it still shows the little "installed" icon. I'm not sure what it is using to compare that, but I assumed that given that the package IDs are different, that would be different as well?

If I enable source support again this issue should fix itself. 

> I noticed this in the commit log; is there an implementation to think about? (it's been shifted to SettingPage`.xaml.cs:441`)

This might me my fault, VS might have created it automatically and then I used a lambda. I will get this fixed.

mrixner:
> What I will do is I will add source compatibility again, but I will manually remove the Source Manager UI from the settings window.

Sounds good, thank you!

> This can be kinda fixed. On the LoadManager() method, you can check if the setting is empty, and if it is not set, it can be set to a default value. I don't know how to get the system default triplet, but if you know how to and you want to fix this it'd be great,

Yes, I noticed and realized this as I was looking through the settings to add a restart required banner for #2883. I'll go do that in a second.

> If I enable source support again this issue should fix itself.

Ahh, of course. That makes sense.

mrixner:
> If I enable source support again this issue should fix itself.

Unfortunately, it doesn't seem to have fixed itself.

image
image

Everything else is looking good, though!

marticliment:
> **CORRECTION**: this doesn't work when an item was just uninstalled, but it does if it's not installed when UniGetUI starts. Therefore, this is not an issue.

This is a bug I am aware of, uninstalled packages still show as installed on the discover page

src/UniGetUI.PackageEngine.Managers.Pip/Pip.cs
```diff
@@ -18,6 +19,17 @@ public class Pip : PackageManager
 
         public Pip()
         {
+            Dependencies = [
+                // GIT is required for vcpkg updates to work
```
mrixner: This comment annotates this wrong; shouldn't it be in reference to Pip instead?

marticliment:
I will merge once I add the icon. The rest looks fine for me

mrixner:
> I will merge once I add the icon. The rest looks fine for me

OK, thank you! Do you need me to make the icon, or will you do that yourself?

mrixner:
Also, just a note in case any issues are opened about vcpkg package discovery taking a while - I have tested this by running `vcpkg search` from the command line, and the first `search` / `list` operation every once in a while just takes a rather long time; this isn't an issue with UniGetUI.

marticliment:
> OK, thank you! Do you need me to make the icon, or will you do that yourself?

I'll do this myself

marticliment:
Vcpkg icon
image
image

marticliment (approved):

mrixner:
> Vcpkg icon image image

Looks great!

marticliment (approved):

PR #2926, opened 2024-11-03
Fix Notification Text Initialization

CoolSpy3:
-----

Currently, there is a bug where the notifications for updating packages (and likely related operations) have the text "THIS NEEDS TO BE REDEFINED ON THE CONSTRUCTOR". After some digging, it looks like this is because `INSTALLING_STRING` is set in the constructors of `AbstractOperation` implementations, but `AbstractOperation` tasks are started asynchronously by the `PackageOperation` and `SourceOperation` constructors. This means that the tasks are started before the subclass constructor is run, leading to a race condition. This PR moves the string initialization in all the places I could find it to the `Initialize()` function, which is always run before the task.

I was able to successfully test this for package updates and removing a manager source. However, because I am not familiar enough with the codebase to setup a reliable way to show an arbitrary notification, and the actual notifications disappear fairly quickly, it is difficult for me to run more in-depth tests on all cases.

-----

I didn't find an open issue for this, but if there is one, feel free to link it!


marticliment:
I don't think there was an issue for it, but I recall having experienced this bug once or twice...

Thanks for the fix! I will have your PR merged in 1 or 2 weeks approx.

marticliment (approved):

PR #2994, opened 2024-11-18
Add vcpkg Package Details

mrixner:
-----

This pull request adds the little package information available on the vcpkg database to UniGetUI.


marticliment:
I removed the tags loaded from the compatible platforms because on most packages it would show a weird expression stating platform compatibility for certain parts of the library

marticliment (approved):

mrixner:
That was the idea, to show if the library would be compatible with your current triplet, but it's also not really a tag so that makes sense.

Issue #1217, opened 2023-07-15
[FEATURE REQUEST] Discover Packages – Search: Browse the search results by pressing the down or up arrow keys


RavenMacDaddy:
### Please confirm these before moving forward

- [X] I have searched for my feature proposal and not found a work-in-progress/duplicate/resolved/discarded issue.
- [X] This proposal is a completely new feature. If you want to suggest an improvement or an enhancement, please use [this template](https://github.com/marticliment/WingetUI/issues/new?assignees=marticliment&labels=enhancement&projects=&template=enhancement-improvement.yml&title=%5BENHANCEMENT%5D+%28Enter+your+description+here%29)

### Describe the new feature

That's, after typing the query, pressing the down arrow key would select the top result, and continuing to press the down arrow key, you continue to go down the list.

Pressing the up arrow key selects from the bottom and upwards.

Obviously, the ability to wrap around would be appreciated; If you're at the bottom entry and press the down arrow key, wrap around to the top entry of the results, and vice versa.

Once on the desired entry, press ENTER to execute.

---

Also, pressing ENTER on the selected entry in the list (i.e., app) ought to respect the option to instantly install by double-clicking;

Option off: Shows details of the selected app when pressing ENTER.
Option on: Starts installation of the selected app when pressing ENTER.

### Describe how this new feature could help users.

More effective navigation, never leaving the keyboard to manage WingetUI.

mrixner:
Unfortunately, this is not possible in a useful manner; there's no way (that we've found) to manually change the focus of the list's items, meaning all the selection would do is move the blue outline rather than actually changing the focus (so none of the other keyboard commands, like the enter key, would work). UniGetUI currently has (the default behavior) something similar; it doesn't wrap around, but it _does_ allow other keyboard interaction, so I think we have to stick with that.
See #3018.

PR #3018, opened 2024-11-26
Better keyboard navigation

mrixner:
-----

This pull request allows typing in a package list view to jump to that package, much like how Windows Explorer allows you to type out a file name on your keyboard outside of the search bar to jump directly to the file.

I tried to solve #1217 but I couldn't get `SelectedIndex` and such to work, so I ended up using `FilteredPackage.IndexOf(PackageList.SelectedItem)` and then after all of that apparently the `Select` call doesn't _actually_ select the element and pressing enter would open the package details for the previously selected element, which I don't know how to fix (or even what the problem is). It occurs to me that this affects this code as well, but the main purpose is to focus the element on the screen.


marticliment:
> I tried to solve #1217 but I couldn't get `SelectedIndex` and such to work, so I ended up using `FilteredPackage.IndexOf(PackageList.SelectedItem)`

You can fix this by using the `Index` property of the `PackageWrapper` items that come from the `ObservablePackageCollection` structure (`AbstractPackagesPage.FilteredPackages`), instead of finding the package on the list (with still would work but wouldn't be as efficient.)


I think #1217 is more of a limitation with how ItemsView handles item focus by default. In fact, after pressing a package on the list arrow navigation works.

Perhaps I could implement arrow press detection, and if the user presses an arrow key and no item is selected the first/last item could be manually selected and then focus given to the package list, so it feels more like the Windows explorer. If you want, I can do this.

mrixner:
> You can fix this by using the Index property of the PackageWrapper items that come from the ObservablePackageCollection structure (AbstractPackagesPage.chat), instead of finding the package on the list (with still would work but wouldn't be as efficient.)

Ahh, OK. Done, thanks.

> I think https://github.com/marticliment/UniGetUI/issues/1217 is more of a limitation with how ItemsView handles item focus by default. In fact, after pressing a package on the list arrow navigation works.

Yeah, I noticed this. #1217 was, I think, primarily about making it loop back around, but I've also noticed that occasionally using the arrow keys moves the ListView, not the focused element.

> Perhaps I could implement arrow press detection, and if the user presses an arrow key and no item is selected the first/last item could be manually selected and then focus given to the package list, so it feels more like the Windows explorer. If you want, I can do this.

Yeah, this isn't that hard; fixed in 1d1ee33.

I think the two solutions to making the keyboard navigation actually useful is that `SelectAndScrollTo` needs to put the focus on the selected element, which I couldn't figure out how to do, or `PackageItemContainer_KeyUp` ([AbstractPackagesPage.xaml.cs:907](https://github.com/mrixner/UniGetUI/blob/fee02f95a450ac16b5eec36398afc5db84a560f9/src/UniGetUI/Pages/SoftwarePages/AbstractPackagesPage.xaml.cs#L909)) needs to use the `SelectedItem` instead of the `sender`.

marticliment:
> I think the two solutions to making the keyboard navigation actually useful is that `SelectAndScrollTo` needs to put the focus on the selected element, which I couldn't figure out how to do, or `PackageItemContainer_KeyUp` ([AbstractPackagesPage.xaml.cs:907](https://github.com/mrixner/UniGetUI/blob/fee02f95a450ac16b5eec36398afc5db84a560f9/src/UniGetUI/Pages/SoftwarePages/AbstractPackagesPage.xaml.cs#L909)) needs to use the `SelectedItem` instead of the `sender`.

I don't think the second is possible, do you think that forcing focus on the ItemsView on SelectAnScrollTo could fix the issue?

mrixner:
> I don't think the second is possible, do you think that forcing focus on the ItemsView on SelectAnScrollTo could fix the issue?

Yeah, I'm pretty sure the problem is that while the item is selected, the focused item doesn't change, and the focused item is the one sending the event.

mrixner:
Correction: forcing focus on the _ItemView_ wouldn't work, the focus has to be on the element being selected.

marticliment:
I am then afraid that this issue cannot be fixed by a limitation on WinUI

mrixner:
Ahh, that's unfortunate. I would've really appreciated this :( I suppose I should remove the arrow key code then, since it doesn't make much sense to a) hinder the default behavior and b) have arrow key selection that doesn't actually allow enter-for-package details. Is my assessment correct?

mrixner:
> I don't think the second is possible

Could I ask why not? Does the Sender expose something `SelectedItem` doesn't? Or is it better practice?

marticliment:
> Could I ask why not? Does the Sender expose something `SelectedItem` doesn't? Or is it better practice

Tested it twice and didn't succeed. They seem to be unreliable. Or at least when I tried I couldn't get them to work.

marticliment:
I have been testing the code, and I have observed the following behaviour:
Pressing multiple times the same key does not iterate over packages starting with the same key (it should do that if the goal is to mimic windows explorer). Instead, it jumps to a random package. Let me give details:
 1. Press letter `b`. Package _BDlot DVD master_ gets highlighted (as expected).
 2. Press letter `b` (again). Instead of _Beaker_ being selected, _Anki_ gets selected, which is not even the first package that starts with letter a.
 3. Pressing the letter `b` more times will remain on _Anki_ being selected.

marticliment (changes requested):
Issue mentioned on the PR conversation

mrixner:
Anki is probably being selected because it's ID has a `bb` in it? I noticed this in explorer the other day but I didn't even know that was the functionality before lol... I'll go fix it.

marticliment:
> Anki is probably being selected because it's ID has a `bb` in it?

No, it's id is `Anki.Anki`

mrixner:
Well that's really bizarre then, I can't imagine why that would be.

marticliment:
Please mark the conversation as resolved when you can fix the issue, thanks!

mrixner (commented):
Fixes implemented

mrixner:
> Please mark the conversation as resolved when you can fix the issue, thanks!

Uhh... I'm not sure how...

marticliment (approved):

PR #2985, opened 2024-11-16
Allow Desktop Shortcut Deletion

mrixner:
- [x] **I have read the [contributing guidelines](https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#coding), and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**.
- [x] **Have you checked that there aren't other open [pull requests](https://github.com/marticliment/WingetUI/pulls) for the same changes?**
- [x] **Have you tested that the committed code can be executed without errors?**
- [x] **This PR is not composed of garbage changes used to farm GitHub activity to enter potential Crypto AirDrops.**
Any user suspected of farming GitHub activity with crypto purposes will get banned. Submitting broken code wastes the contributors' time, who have to spend their free time reviewing, fixing, and testing code that does not even compile breaks other features, or does not introduce any useful changes. I appreciate your understanding.
-----

This pull request allows users to have UniGetUI delete shortcuts automatically added to their desktop. This will be off by default, behind a setting, and require explicit user confirmation before UniGetUI deletes anything.

- [x] Change desktop shortcut storage to use the new complex settings
- [ ] Copy the detection code to the `InstallPackageOperation` as well and have a layer setting ([marticliment](https://github.com/marticliment/UniGetUI/pull/2985#discussion_r1860082539))

-----

Closes #2942
Closes #2080 
Closes #624 
Relates to #2160

PR #3025, opened 2024-11-28
Complex Settings

mrixner:
-----

This PR adds complex setting support to the settings engine (lists, dictionaries). 

Things I'm planning to accomplish:
- [x] Add List Functions
- [x] Add List Tests
- [x] Add Dictionary Functions
- [x] Add Dictionary Tests
- [x] Make Sure Exporting These Settings Works Properly
- [x] Transfer Ignored Updates (Dictionary)
- [x] Code To Automatically Port Old Ignored Updates

I know we discussed using `ISerializable` but despite extensive searching I was unable to find a way to easily, automatically serialize `ISerializable` to JSON because apparently C#'s JSON thing doesn't support `ISerializable`. Therefore, I'm just converting everything to `object`s, casting them around and letting the JSON library serialize how it wants.


marticliment:
Yes, this looks good for me.
Perhaps you could (in order to not have a huge file with the entire settings code) mark the settings class as `partial`, and have List-related and Dictionary-related settings be implemented in their own files. You can find an example on what I mean on src/UniGetUI/Pages/DialogPages/DuialogHelper_*.cs, which are different files that define different parts of the same static class

marticliment:
> Actually, unless you would like me to port them, I think I'll just leave the installation options where they are because the current storage method because it's pretty ingrained. But I'd be happy to port them if you want consistency.

I'd leave them as they are, as the method use is very customized for the use cases of the settings.

marticliment:
> Do you want the installation options to continue to not be backed up in the settings? Because if so I should update `IgnoredSettings` ([SettingsPage.xaml.cs:537](https://github.com/mrixner/UniGetUI/blob/86e5ce86a1d172c0abcfa92b738d9f2585c23dbe/src/UniGetUI/Pages/SettingsPage.xaml.cs#L537)) to include the installation options. I think probably, since if settings are used to transfer UniGetUI configurations across computers, installation options would be different across computers.

I reckon it makes sense that installation options are exported via package bundles, rather than on the settings

marticliment:
> * Transfer Desktop Shortcuts (Requires [Allow Desktop Shortcut Deletion #2985](https://github.com/marticliment/UniGetUI/pull/2985) to be merged) (Dictionary)
>   (this won't need transferring if I can add it before beta2)

I will merge both of the PRs before beta2, so there will not be any need to migrate any settings.
I reckon it would be a good idea to first merge this PR, so then you can pull main from the "remove desktop icons" PR, what do you think?

mrixner:
Yeah, sounds good! Thanks.

marticliment (approved):

marticliment (approved):

marticliment:
I did, because if the tests failed and test files remained the tests would fail infinitely

PR #3101, opened 2024-12-20
Skip minor updates

### LINKED ISSUE #2927: [FEATURE REQUEST] Add option per package to only skip minor versions

- [X] This proposal is a completely new feature. If you want to suggest an improvement or an enhancement, please use [this template](https://github.com/marticliment/WingetUI/issues/new?assignees=marticliment&labels=enhancement&projects=&template=enhancement-improvement.yml&title=%5BENHANCEMENT%5D+%28Enter+your+description+here%29).

### Describe the new feature

An option in the `Software Updates` view that lets users choose to skip minor versions of some packages.
Those packages don't show the latest update available if it's a minor version update. (just like `Skip this version`)

This feature could be even upgraded to support semantic versioning: MAJOR.MINOR.PATCH.BUILD

I would expect the context menu of a package to have an option to select which version levels are relevant to me.
Default would be `BUILD` which may confuse users, so maybe `ALL`?

### Describe how this new feature could help users

This feature would solve an issue with packages that have daily or weekly updates available, that aren't relavant to some users.

Ignoring the package completely is not the best way to go here and skipping each version is still manual work.

So beeing able to select the level of version the user wants to stay current, is a fitting solution for this issue

mrixner:
-----

This PR makes the user able to skip minor updates for a package in the installation options. Minor updates are determined as the "z" portion (or "z#w" in the case of vcpkg) of two versions "x.y.z{#w}" differing. If "x" or "y" differ, it will not be counted as a minor upgrade, as well as if there aren't three parts of a version for both.

This also fixes an issue where vcpkg installation options files would use the wrong name.
image

Closes #2927


marticliment:
Will check later today or tomorrow

src/UniGetUI.PackageEngine.Package/InstallationOptions.cs
```diff
@@ -96,6 +96,7 @@ public void FromSerialized(SerializableInstallationOptions_v1 options)
             RunAsAdministrator = options.RunAsAdministrator;
             CustomInstallLocation = options.CustomInstallLocation;
             Version = options.Version;
+            SkipMinorUpdates = options.SkipMinorUpdates;
```
marticliment: I am not sure if the SerializableInstallationOptions_v1 can be easily changed maintaining back compatibility for files that don't have this new option set; both for importing/exporting packages and for saving/loading package settings

marticliment: I will test it tomorrow more deeply to see if it works

mrixner: Of course you should test it as well, however I thought about this beforehand and explicitly did some testing and it just loaded the new item to the default value without causing any problems.

mrixner: Although, that was just for loading the already existing options files; I don't think I tested exporting / importing stuff, but I don't think it would be different.

marticliment:
I have updated the export_version from 2 to 2.1.

I have been doing some testing, and I have found that import/export work fine, except for the case where a version 2.1 `.yaml` file does not load on UniGetUI 3.1.5 and lower. However `.xml` and `.json` (and `.ubundle`, which is internally a json) seem to work fine. To circumvent the issue, the affected yaml bundles can be easily imported on a newer version and re-exported to Json, so I wouldn't consider this a big deal. Nevertheless, I have tweaked the YAML deserialized so it ignores unknown fields (and this regression does not happen again in the future.)

I would appreciate if you could test this and confirm you are seeing the same behaviour as me

src/UniGetUI.PackageEngine.PackageManagerClasses/Packages/Classes/InstallationOptions.cs
```diff
@@ -166,7 +169,7 @@ public SerializableInstallationOptions_v1 AsSerializable()
 
         private FileInfo GetPackageOptionsFile()
         {
-            string optionsFileName = Package.Manager.Name + "." + Package.Id + ".json";
+            string optionsFileName = Package.Manager.Name + "." + Package.Id.Split(":")[0] + ".json";
```
marticliment, line 172: I suppose this is for vcpkg, right?

marticliment (approved):

src/UniGetUI.PackageEngine.PackageManagerClasses/Packages/Classes/InstallationOptions.cs
mrixner: Yes; it was putting colons in the file name resulting in it being `vcpkg.curl` or whatever as opposed to `vcpkg.curl.json`. It broke consistency and I figured it's never a good idea to do unsupported stuff (using an unsupported character in a filename).

mrixner:
I'm sorry, I'm not totally sure what you're requesting that I test. Just that skipping minor updates works, or serialization / deserialization of the install options for YAML files, or something else?

marticliment:
Testing importing format 2.0 (old format) on this unigetui branch, and testing importing 2.1 format (new format) into UniGetUI stable (3.1.5 or lower). Opening the bundles should be enough, there is no need to install the packages, in the three formats json, xml and yaml. 

Everything should work fine except opening the Yaml format 2.1 on UniGetUI stable.

I have tested it, but I don't want to risk it on such an important feature.

mrixner:
Package bundles: 
image

-------------------------------
**Import Results**: (including with modified installation options that _were_ reflected)
3.1.5 UBundle -> skip-minor-updates: ✔️\*
3.1.5 JSON -> skip-minor-updates: ✔️\*
3.1.5 YAML -> skip-minor-updates: ✔️
3.1.5 XML -> skip-minor-updates: ✔️
skip-minor-updates UBundle -> 3.1.5: ✔️
skip-minor-updates JSON -> 3.1.5: ✔️
skip-minor-updates YAML -> 3.1.5: ❌\*\*
skip-minor-updates XML -> 3.1.5: ✔️

\* [note: did not happen again, cannot reproduce]
image
image
image
It loads the packages, however doesn't show them in the list. It seemed to work until I manually changed `SkipHashCheck` to true, and then even changing it back resulted in this blank packages list. I restarted UniGetUI, which fixed it. I could not reproduce this, thus it's likely not an issue.

\*\*
image
This is the one you were talking about, and I agree that it shouldn't matter as most people wouldn't import a newer bundle into an older version. However, if this does occur, perhaps you could give a warning that the `SerializableInstallationOptions_v1` export version is older than the imported bundles and they should update (if you think it matters), and I think that's a fairly reasonable compatibility restraint. I don't know how much bundles are used but I think the really only scenario this would occur on is if someone had a bundle collection of packages they thought were useful that they shared with others who may then have an old version of UniGetUI.

marticliment:
Not seeing the packages rendered is likely to be https://github.com/marticliment/UniGetUI/issues/2737

> However, if this does occur, perhaps you could give a warning that the `SerializableInstallationOptions_v1` export version is older than the imported bundles and they should update (if you think it matters), and I think that's a fairly reasonable compatibility restraint.

The issue here is that this will only happen on already released UniGetUI versions, so it can't really be fixed/adverted to the user.

mrixner:
> The issue here is that this will only happen on already released UniGetUI versions, so it can't really be fixed/adverted to the user.

You're right, I didn't think of that 🤦

PR #3108, opened 2024-12-20
TRIM Support

marticliment:
Enable trim, decreasing the size of the executable file and decreasing RAM usage.
Step prior to AOT

Normal executable size: 199mb
Trimmed executable mode: 114mb

# TODO 
 - [x] ~~Notifications not working~~ (turned out to be caused by another issue, already fixed)
 - [ ] Expanding a Settings Expander crashes UI
 - [x] Settings test not working


marticliment:
It looks like SettingsExpander is not compatible with trimming yet

PR #3119, opened 2024-12-23
Updated Winget Workflows to Newer Versions

vikingnope:
-----

This update enhances the WinGet release workflow to ensure it processes only the specific installer file, `UniGetUI.Installer.exe`. It also ensures to use the latest version of the `winget-releaser` since the old v2 release still uses the old Manifest Version 1.5.0, which is not encouraged anymore due to it now being on version 1.9.0; this is happening due to the v2 version of 'winget-releaser` using Komac version 1.11.0, currently we are on Komac version 2.8.0.

I think new releases will not happen for `winget-releaser` since the whole codebase seems to have changed from v2 to now and it is now using mostly CLI commands, so I think this is the way forward.

-----

### DIFF
--- a/.github/workflows/winget-prerelease.yml
+++ b/.github/workflows/winget-prerelease.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
       - name: Publish to WinGet
-        uses: vedantmgoyal2009/winget-releaser@main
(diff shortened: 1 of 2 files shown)


marticliment (approved):

marticliment:
Thanks!

PR #3139, opened 2024-12-28
Update update-icons.yaml: Don't run for forks

Saibamen:
Don't spam PRs for forks


marticliment (approved):

marticliment:
My excuses. I fixed this in the past, but only for the translations one

PR #3167, opened 2025-01-07
Update "Pause Updates" to Use ICU

mrixner:
-----

I tried to make the format string automatically handle the stuff, but it doesn't seem like it supports any plural matching aside from specific numbers. Feel free to revert / change anything you prefer the other / a different version of, as always.


marticliment:
I have looked at this, and, to be honest, I reckon it will be more confusing (both for me and for the translators) to have to translate this long strings than to use the already existing strings differentiating one and many.

Furthermore, most of this strings are already translated (since they are already used on the settings page), so I am going to close this pr.

Nevertheless, thanks for putting the effort into it.

PR #3176, opened 2025-01-10
Load translations from Tolgee

martinet101:


Denisskas:
@marticliment Hello, you can check Tolgee? I fix words.

marticliment:
Thanks for the fix! I will run the translations updater again

PR #3189, opened 2025-01-14
Diversify Icon Detection (Icons for Steam, Local PC, and Microsoft Store Applications)

mrixner:
-----

This pull request allows adding icons to apps from Steam, Local PC, or Microsoft Store sources much more easily. I don't have an GOG or Ubisoft Connect apps installed, so I couldn't add cases for them, and the Android Subsystem for Linux I believe has been discontinued so I didn't bother with that either.

Steam IDs resolve to just a bunch of random numbers, and I didn't know if that would fit with the current sheet (after all, what are random numbers supposed to tell you?), so I prefixed it with `steam:` to clarify that it's a Steam app. Local PC I left as it is (removing the extra metadata), because while some are UUIDs some are also actual package names*. Microsoft Store apps I removed the `MSIX\` and publisher (like WinGet's integration does) as well as removing the random character identifier at the end (e.g. `MSIX\Microsoft.OneDriveSync_24226.1110.4.O_neutral_8wekyb3d8bbwe` -> `OneDriveSync`).

\*
image
image

This PR also enables Steam screenshots / metadata via the [internal Steam API](https://github.com/Revadike/InternalSteamWebAPI/wiki). I was unable to find an API call to get the square logos used in the library page, though, so unless one of the rectangular icons the API DOES provide (which if you would like is not currently implemented and would need me to go back and change something) is acceptable, Steam packages will have to rely on user contributions.

**NOTE:** Currently, in the Package Details page, the "download installer" button opens the Steam install dialog as it is supposed to, but also a random Windows system dialog saying "there's no app on your PC that can open this `about:` link", which doesn't make sense since I have no `about:` links being started. The `installer url` field also does not properly open the Steam install dialog when clicked, despite it being the same URL used for the working `Download installer` button, but I think that's intentional, right?
image
image

I tried to put the Steam check in the `IWinGetManagerHelpers` class but I'm not proficient enough with C# to have a virtual interface body method that I can call in an overriding class, so I just duplicated the implementation.


marticliment:
> I tried to put the Steam check in the `IWinGetManagerHelpers` class but I'm not proficient enough with C# to have a virtual interface body method that I can call in an overriding class, so I just duplicated the implementation.

I will see what I can do to not have this code duplicated.

marticliment:
I honestly don't think the steam game banners would work as icons, as they show too much detail.

However, I will see if the .ico file can be easily extracted from a game installation, as the desktop shodtcuts show more suited icons.

mrixner:
> In fact, steam apps shouldn't even show a download URL...

I know; this is something I added in this PR, when I added Steam package details. It's a link to the steam browser protocol (`steam://install/id`), but clicking it doesn't open Steam like "Download installer" does (which it should, since they use the same Steam browser protocol link).

marticliment:
> I know; this is something I added in this PR, when I added Steam package details. It's a link to the steam browser protocol (`steam://install/id`), but clicking it doesn't open Steam like "Download installer" does (which it should, since they use the same Steam browser protocol link).

To be honest, I don't think it is really important to show that download URL, since steam apps only show on the installed packages page, so the user must already have them installed... Steam games won't show on the discover or updates pages.

marticliment:
Furthermore, Local packages (Steam packages are local packages) shouldn't be able to show their package details...

mrixner:
> I honestly don't think the steam game banners would work as icons, as they show too much detail.
> 
> However, I will see if the .ico file can be easily extracted from a game installation, as the desktop shodtcuts show more suited icons.

Yeah, not the banners, but the API I used has a "logo" property; it's not a square and is fairly large, but I think it's _possible_ it could work, but I don't think it fits great, hence why I didn't add it initially.

mrixner:
> Furthermore, Local packages (Steam packages are local packages) shouldn't be able to show their package details... The entry on the context menu should be greyed out, and the toolbar button shouldn't work either...

Again, my apologies. I changed this so it would work for Steam explicitly since I figured I could get the package details, might as well show them. I didn't realize you didn't want that; in that case, just revert the last commit and it _should_ get rid of all the Steam package details while leaving the icon ID detection, which was the original point of this PR before I got sidetracked adding Steam details.

mrixner:
> I didn't realize you didn't want that; in that case, just revert the last commit and it _should_ get rid of all the Steam package details while leaving the icon ID detection, which was the original point of this PR before I got sidetracked adding Steam details. 

Sorry, all of the commits besides the first*. Let me know if you want me to remove it, I'll open a new PR with just the icon changes when I can.

mrixner:
Done

marticliment (approved):

PR #3247, opened 2025-02-03
Update update-tolgee.yml: Run every Monday

Saibamen:
Reduce git pull requests and commits

### DIFF
--- a/.github/workflows/update-tolgee.yml
+++ b/.github/workflows/update-tolgee.yml
@@ -2,7 +2,7 @@ name: Tolgee CI
 
 on:
   schedule:
-    - cron: 0 0 * * *
+    - cron: 0 0 * * 1
   workflow_dispatch:
 jobs:
   load-translations:


marticliment:
Why?

Saibamen:
To reduce pull requests you need to merge and commits in changelog view.

Right now there are 261 commits between 3.1.5 and 3.1.6-beta3: https://github.com/marticliment/UniGetUI/compare/3.1.5...3.1.6-beta3

And 29 of them are for translations. Lot of them are only updating one or two languages.

marticliment:
> And 30 of them are for translations. Lot of them are only updating one or two languages.

Yes, but updated translations take effect on already published versions of UniGetUI. Therefore, by decreasing the amount of PRs the changes would take much longer to be effective

Saibamen:
>Once a PR is merged, UniGetUI will pull new translations from GitHub

Ahh ok. This is not included in UnigetUI source code, but downloaded from GH and cached on local disk.

marticliment:
Yep, exactly

PR #3344, opened 2025-02-22
Package list: ignore modifier key press

tkohlmeier:
-----

Currently, pressing Control, Shift or Alt key results in selecting the first item in a package list. This PR fixes this by only handling `[a-z0-9]` keys.


marticliment:
Will test and merge later

tkohlmeier:
Found an even better solution. The `CharacterReceived` event seems to be exactly what is needed here.

mrixner:
Yeah, I kind of patched this together. Thank you got improving it!

marticliment (approved):

PR #3368, opened 2025-02-27
Add package rankings

marticliment:
image


theguy000:
Why is this not on the main app?

marticliment:
Had issues with the statistic server, still need to figure some stuff out

PR #3705, opened 2025-05-31
Make 'Pause updates for' submenu item use translation

szumsky:
- [X] **I have read the [contributing guidelines](https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#coding), and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**.
- [X] **This PR is not composed of garbage changes used to farm GitHub activity to enter potential Crypto AirDrops.**
-----
This submenu item was not translated and was missing the call to 'CoreTools.AutoTranslated'.
After rebuilding the item was still not translated - not sure about this, but it seems that you need to use the tools from the scripts directory to update the translations, which I couldn't do, because they required an API key and the one I generated myself on tolgee didn't work (403).

### DIFF
--- a/src/UniGetUI/Pages/SoftwarePages/SoftwareUpdatesPage.cs
+++ b/src/UniGetUI/Pages/SoftwarePages/SoftwareUpdatesPage.cs
@@ -153,7 +153,7 @@ namespace UniGetUI.Interface.SoftwarePages
 
             MenuFlyoutSubItem menuPause = new()
             {
-                Text = "Pause updates for",
+                Text = CoreTools.AutoTranslated("Pause updates for"),
                 Icon = new FontIcon { Glyph = "\uE769" },
             };
             foreach (IgnoredUpdatesDatabase.PauseTime menuTime in new List<IgnoredUpdatesDatabase.PauseTime>{


src/UniGetUI/Pages/SoftwarePages/SoftwareUpdatesPage.cs
```diff
@@ -153,7 +153,7 @@ public override BetterMenu GenerateContextMenu()
 
             MenuFlyoutSubItem menuPause = new()
             {
-                Text = "Pause updates for",
+                Text = CoreTools.AutoTranslated("Pause updates for"),
```
marticliment: `CoreTools.AutoTranslate` should only be used when the field automatically translates the text (this happens with `BetterMenuItem`, but not with the native `MenuFlyoutSubItem` class, and is used to warn a custom script I have that this is a translatable string). In this case, `CoreTools.Translate` should be used instead.

marticliment (changes requested):

src/UniGetUI/Pages/SoftwarePages/SoftwareUpdatesPage.cs
szumsky: My bad, changed.

marticliment (approved):

Issue #3038, opened 2024-11-30
[FEATURE REQUEST] Run command before updating.


Goth667:
### Please confirm these before moving forward.

- [X] I have searched for my feature proposal and have not found a work-in-progress/duplicate/resolved/discarded issue.
- [X] This proposal is a completely new feature. If you want to suggest an improvement or an enhancement, please use [this template](https://github.com/marticliment/WingetUI/issues/new?assignees=marticliment&labels=enhancement&projects=&template=enhancement-improvement.yml&title=%5BENHANCEMENT%5D+%28Enter+your+description+here%29).

### Describe the new feature

Hi,
I want do propose a feature in the installation options of a package. It would be nice if it was possible to run a command before the installation or update. 
A simple text field where you can add the command you want executed and a checkbox if it should run as Administrator. 
Thanks

### Describe how this new feature could help users

Some packages can't automatically be installed if the program is already running, hanging up or its files are already in use by another application with a shared library. Depending on the system, it would be possible to add a command that shuts down the application or deletes the settings file etc.
![image](https://github.com/user-attachments/assets/d04292bc-9bd2-4dee-b64d-05e1644fcef0)

mrixner:
Duplicate of #2160 and #2193

Goth667:
@mrixner 
Sry, apparently I did a bad job a searching before posting.

@marticliment 
Would adding a user for impersonation to run the command work?

Goth667:
Thanks for the explanation.

So having to set up a user under whom the command will be executed would be harder to do then to double click a .reg file. 
Should we also limit it so the impersonated user can't be the current user?
A warning pop up would also help, I guess.

In the end if you get the user to run a .reg or script there is no need to do it in UniGetUI.

In the end it boils down to the war against the universe. Where as devs try to make software safer and safer for dumb users and the universe trying to produce dumber and dumber users.

mrixner:
One method I could think of is simply asking the user for confirmation about every command; every time it tries to run a post-install command, UniGetUI will check if that _specific command_ has been approved before, and if not throw up an approval dialog. This way, an externally modified command will need approval first. And to prevent the external command from just being set as approved, maybe there could be a "hook password" in the settings used to encrypt the list of approved commands?

marticliment:
> UniGetUI will check if that _specific command_ has been approved before, and if not throw up an approval dialog.

[...]

PR #3722, opened 2025-06-07
Add toggle to enable/disable insecure settings

marticliment:
Certain features, such as custom command-line arguments, and future ones such as pre-install and post-install scripts, pose an important security risk. They should be disabled by default, and in a way that a remote, unauthorized agent cannot manually enable this features.

 - [ ] Implement _secure settings_ (can only be changed with administrator permissions, yet they should be user-based)
 - [ ] Create a settings page to handle these secure settings. Create a UI warning to enable and disable them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ The main goal of this project is to create an intuitive GUI for the most common 
 With this app, you can easily download, install, update, and uninstall any software published on the supported package managers — and much more!
 
 ![image](https://github.com/user-attachments/assets/7cb447ca-ee8b-4bce-8561-b9332fb0139a)
-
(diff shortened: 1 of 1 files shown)


mrixner:
Does this enable the implementation of #3038 as well? Or do you still have reservations about that?

marticliment:
Yeah, it does. However, I want to do a generic pre and post install steps, to allow for more things aside from custom commands (kill an app before updating, cleanup after update, etc.)

mrixner:
OK, I'll leave that one to you then, but I'm happy to help if you would like.

marticliment:
Wait this is not supposed to be closed

PR #3750, opened 2025-06-15
Setting keys will be stored on const strings

marticliment:
For easier coding, settings keys will be stored on ~~const strings~~ an enumeration


griush:
estudia coño, tot el dia fent PRs

marticliment:
que pesao jajajajaj ves a estudiar BD

PR #3906, opened 2025-07-22
Migrate to MSIX [guarantee package integrity]

marticliment:
Release 3.3.0 experiences an issue where, for unknown reasons, some files become corrupted or missing, causing UniGetUI to crash. This issue, in reality has always happened (issues with AntiViruses, etc.). However, with this latest release, the issue has become worse. This is why, UniGetUI will ditch the regular `C:\Program Files\UniGetUI` install, in favour of a msix package. 

Notes on this:
 - _No sacrifices will be made_: All working features will continue working. (or at least this is the idea)
 - UniGetUI will still be able to run outside of a package: `UniGetUI.x64.zip` will contain the executables
 - Uninstalling UniGetUI will be easier with MSIX packages
 - The _regular_ installer will take care of uninstalling the older version, and deploy the new msix package.
 - The update process will be (almost) unnoticeable. The only change may be the disapparition of the tarkbar icon, and autostart on system launch setting may get reset.
 - This will be tested thoroughly before being deployed into production.


marticliment:
It looks like it is impossible to call `cmd.exe` from an elevated from a MSIX package :(

PR #4107, opened 2025-09-12
Clamp filters column width, collapse earlier

Simek:

> [!IMPORTANT]  
> Hello Martí, thank you so much for creating this tool! Since I have stumbled upon it one day, I cannot imagine managing installed packages and apps on Windows without it! 🙏 

# Why

I have played with UniGet UI one day and resized the filters pane over the maximum grid container width which due to storing the last set value in setting made the one of the view unusable for me (reinstall/update did not help, so I decided to build app from the source and adjust ActualWidth value via Property Explorer, which inspired that PR).

I have also spotted that width at which filters pane collapses could be higher, since below 200 pixels buttons and radios labels break line or become trimmed. Under ~150 pixels the content becomes almost fully overlapped via container spacing and other elements, so is seems reasonable to collapse filters pane earlier.

# How

The following PR introduces the following changes:
* sets max width of filters pane to 600 pixels via value clamping in `onSizeChanged` listener to prevent situation like on the recording above
* changes the width at which filters pane collapses from 100 pixels to 204 pixels, to make sure that content is fully visible in minimal allowed pane width

I'm happy to adjust the values if you think that other ones would work better, especially 600 pixels max width is a quite arbitrary choice, and maybe it could be set to even lower value.

# Preview

### DIFF
--- a/src/UniGetUI/Pages/SoftwarePages/AbstractPackagesPage.xaml.cs
+++ b/src/UniGetUI/Pages/SoftwarePages/AbstractPackagesPage.xaml.cs
@@ -1067,7 +1067,7 @@ namespace UniGetUI.Interface
 
             int newWidth = Math.Clamp(rawWidth, 0, (int)this.ActualWidth - 250);
 
-            if (newWidth < 204)
+            if (newWidth < 100)
             {
                 HideFilteringPane();
                 Settings.SetDictionaryItem(Settings.K.SidepanelWidths, PAGE_NAME, 250);


Simek:
Thanks for the quick review and merge @marticliment! 👍 

However, after testing the code with applied changes locally, it looks like the filters pane clamping did not work anymore, and situation like described above (with filters pane being extended over the container width) still can happen.

marticliment:
The limit is set to the page width - 250px, for me it works

marticliment:
You are right, it works better with the 600

Simek:
Okay, I see what's the issue. It works correctly when I reduce the window size so the app container does not have horizontal padding and max width of content is not reached, but with maximized window on 4k display (and with spacing around content is present, in other words max width of 2000 pixels is reached) the altered code allows to resize the filter pane over the container size. 

Screenshot 2025-09-12 222952

I guess that wrapping the ActualWidth with Math.Min against max content size reduced by desired minimal packages list width would do the trick? Just quickly tested the rough change like below, and it seems to work:
```cs
int newWidth = Math.Clamp(rawWidth, 0, Math.Min((int)this.ActualWidth - 250, 2000 - 250));
```

PR #4176, opened 2025-10-12
Startup speed improvement

Taron-art:
-----
There are two related changes:
1. LoadComponentsAsync() was waiting 700 ms (in DoEntryTextAnimationAsync) before starting the iniTasks creation. I moved DoEntryTextAnimationAsync to iniTasks, this helps to shave this 700 ms from start time (25 - 30 % of startup time on my machine).
2. Second commit changed LoadingIndicator default visibility to "Collapsed", as it was probably intended. Otherwise, pause and Visibility set in LoadComponentsAsync does not make sense.
-----


src/UniGetUI/MainWindow.xaml
```diff
@@ -188,7 +188,7 @@
                     VerticalAlignment="Bottom"
                     Foreground="#08a9c3"
                     IsIndeterminate="True"
-                    Visibility="Visible" />
+                    Visibility="Collapsed" />
```
marticliment, line 191: Does this change anything?

src/UniGetUI/App.xaml.cs
```diff
@@ -299,16 +299,16 @@ private async Task LoadComponentsAsync()
 
                 // Create MainWindow
                 InitializeMainWindow();
-                await MainWindow.DoEntryTextAnimationAsync();
 
                 IEnumerable<Task> iniTasks =
                 [
+                    Task.Run(PEInterface.LoadManagers), // Takes most of the time
+                    MainWindow.DoEntryTextAnimationAsync(),
```
marticliment: Running UI things via Task.WhenAll can produde deadlocks. It would be better to have it as `_ = MainWindow.DoEntryTextAnimationAsync();` where it was before

src/UniGetUI/MainWindow.xaml
Taron-art: Yes, the progess bar appears after the animation has ended. You planned this way? I think that because of the Task.Delay(700) in `DoEntryTextAnimationAsync()`. It looks nicer, but if you OK with current behavior, the DoEntryTextAnimationAsync can be made sync and this delay (with Visibility set) can be removed...

marticliment: Oh yes, it was planned. Because otherwhise it looks weird, at least IMO

src/UniGetUI/App.xaml.cs
Taron-art: Because we are doing await there, it shouldn't be an issue. But fire are forget will probably work here without any issues. Please let me know if you want me to change that (depends on first comment answer).

marticliment: Yes, please do. I have had issues in the past

Taron-art: Done.

PR #4396, opened 2026-02-06
Add support for bun (shares most code with Npm)

### LINKED ISSUE #2519: [FEATURE REQUEST] (Add bun support)

- [X] This proposal is a completely new feature. If you want to suggest an improvement or an enhancement, please use [this template](https://github.com/marticliment/WingetUI/issues/new?assignees=marticliment&labels=enhancement&projects=&template=enhancement-improvement.yml&title=%5BENHANCEMENT%5D+%28Enter+your+description+here%29).

### Describe the new feature

Bun is a JS runtime that claims to be much faster than npm. Allow install from bun, which uses the same registry as npm by default.

### Describe how this new feature could help users

Allows installing the same CLI tools that npm can, with the possibility of much faster execution speed.

clin1234:
- [X] **I have read the [contributing guidelines](https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#coding), and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**.
- [X] **This PR is not composed of garbage changes used to farm GitHub activity to enter potential Crypto AirDrops.**

Closes #2519


mrixner:
Isn't Bun drop-in compatible with NPM? We could just add it as an npm executable option.

clin1234:
> Isn't Bun drop-in compatible with NPM? We could just add it as an npm executable option.

Theoretically yes, but I did not know where the input box for custom `npm` executable path is in the codebase.

mrixner:
In Npm.cs there's a `LoadAvailablePaths` method that returns a list of available paths. You can check for bun in there and return the paths there and it will become an option.

clin1234:
I can't find your mentioned path anywhere in upstream: https://github.com/search?q=repo%3Amarticliment%2FUniGetUI+LoadAvailablePaths&type=code

mrixner:
Apologies, the name has changed.

https://github.com/marticliment/UniGetUI/blob/386228d80387f05cd994f5d61b707e1bec39d4e1/src/UniGetUI.PackageEngine.Managers.Npm/Npm.cs#L191

clin1234:
> Apologies, the name has changed.
> 
> https://github.com/marticliment/UniGetUI/blob/386228d80387f05cd994f5d61b707e1bec39d4e1/src/UniGetUI.PackageEngine.Managers.Npm/Npm.cs#L191

Powershell 5 and Powershell 7 have separate `PackageEngine.Managers`, so why shouldn't `bun` be added through a separate Manager, even though the two versions are virtually identical in syntax?

mrixner:
The powershell versions were added before this was implemented, and I believe they have different syntax. Nonetheless, I'm not saying to do that instead of your current solution, I'm just pointing out an alternative approach to this problem in case Martí decides it better. Both have strengths and weaknesses.

clin1234:
@marticliment Care to weigh in on this debate?

marticliment:
sorry, I forgot to answer. I have not used bun enough to have a formed opinion on this. However,

If bun is a drop-in replacement for npm (exact same install/update/uninstall commands, return codes, output structure, etc.) and it manages the same packages (global and user-scoped) then I guess the exe can be added as an alternative to Npm.

However, if there is any difference (cli output, behaviour, etc.) then I'd just add Bun as a separate package manager.

mrixner:
In terms of command structure, if I recall correctly it is supposed to be 1:1. However, while it's the same for project packages, I believe it installs global packages to a different place than NPM? So it might be better as a seperate package manager than as an NPM executable. I just wanted to weigh in with a potential alternative solution.

mamoreau-devolutions:
this pull request is dated - lots have changed in UniGetUI since then. Can you submit an issue instead of a pull request so we can re-evaluate when and how to add support for bun?

clin1234:
> this pull request is dated - lots have changed in UniGetUI since then. Can you submit an issue instead of a pull request so we can re-evaluate when and how to add support for bun?

See #2519

PR #4450, opened 2026-03-11
Run `dotnet format`

Saibamen:
I think right now is the good time to do this, before you started working on C# code itself to reduce merge conflicts in the future.


mamoreau-devolutions:
I'll consider doing a widespread reformatting, but I've got a lot of other cleanup tasks to be done before this one. I would also like to take the time to correctly enforce the formatting checks once it's done. I'd rather have an open issue to track this than a pull request since because it costs nothing to re-run the formatter once we're ready to make the change.

mamoreau-devolutions:
@Saibamen https://github.com/Devolutions/UniGetUI/pull/4459

PR #4517, opened 2026-03-23
docs: correct flag for Afrikaans

JulianPritchard:
The current flag for Afrikaans uses the `af` country code which is Afghanistan, switched to `za` for South Africa

-----

Afrikaans as an official language of [South Africa](https://en.wikipedia.org/wiki/Languages_of_South_Africa) and should use the ZA flag code. Currently the flag code is AF for [Afghanistan](https://en.wikipedia.org/wiki/Languages_of_Afghanistan), which has the official languages of Pashto and Dari.

-----


mamoreau-devolutions:
sorry about that, and thanks for fixing it.

PR #4582, opened 2026-04-14
 Avalonia: add screen reader support, keyboard navigation, and focus management 

GabrielDuf:
This branch adds a comprehensive accessibility layer to the Avalonia UI port:

### Screen reader support (VoiceOver / NVDA / JAWS)
  - Add `AutomationProperties` semantics across the shell, sidebar, package pages, banners, dialogs, and settings cards — names, roles, accelerator keys, labeled-by associations
  - Mark decorative icons and chrome as `AccessibilityView="Raw"` to hide them from the tree
  - Add a shared `AccessibilityAnnouncementService` (live-region) that announces page navigation, operation progress/success/failure, settings changes, manager toggle state, select-all state, and ignore-updates changes independently of platform toast notifications

### Keyboard navigation & shortcuts
  - `Ctrl+1–6` jump to the six main sections
  - `Ctrl+D` opens package details; `F1` opens Help; `F5 / Ctrl+R` reload; `Ctrl+A` select all


src/UniGetUI.Avalonia/ViewModels/MainWindowViewModel.cs
```diff
@@ -203,6 +212,15 @@ public MainWindowViewModel()
         LoadDefaultPage();
     }
 
+    private void OnAnnouncementRequested(object? _, AccessibilityAnnouncement announcement)
+    {
+        AnnouncementLiveSetting = announcement.LiveSetting;
+        AnnouncementText = string.Empty;
+        Dispatcher.UIThread.Post(
+            () => AnnouncementText = announcement.Message,
+            DispatcherPriority.Background);
```
MaxBuisson, line 221: Why set empty and then set message 🤔

src/UniGetUI.Avalonia/Views/Controls/Settings/SecureCheckboxCard.cs
```diff
@@ -149,6 +150,10 @@ await SecureSettings.TrySet(
                 cmd.Execute(null);
             _textblock.Opacity = (_checkbox.IsChecked ?? false) ? 1 : 0.7;
             _checkbox.IsChecked = SecureSettings.Get(setting_name) ^ IS_INVERTED ^ ForceInversion;
+            AccessibilityAnnouncementService.Announce(
+                CoreTools.Translate("{0} is now {1}", _textblock.Text, (_checkbox.IsChecked ?? false)
+                    ? CoreTools.Translate("Enabled")
+                    : CoreTools.Translate("Disabled")));
```
MaxBuisson: This pattern to annonce the state of a checkbox is copied and pasted many times already, should we make a helper function or something ?

src/UniGetUI.Avalonia/Views/DialogPages/InstallOptionsWindow.axaml
```diff
@@ -9,6 +10,10 @@
         Background="{DynamicResource AppDialogBackground}"
         Title="{Binding DialogTitle}">
 
+    <Window.KeyBindings>
+        <KeyBinding Gesture="Ctrl+S" Command="{Binding SaveCommand}"/>
```
MaxBuisson, line 14: Will that work on macOS with Cmd+S ?

MaxBuisson (dismissed):
Looks good overall !

src/UniGetUI.Avalonia/ViewModels/MainWindowViewModel.cs
GabrielDuf: If the new announcement message is identical to the current AnnouncementText, the property change         notification won't fire (since the value didn't change), and screen readers won't re-announce it.

src/UniGetUI.Avalonia/Views/Controls/Settings/SecureCheckboxCard.cs
GabrielDuf: good idea, i'm adding it

src/UniGetUI.Avalonia/Views/DialogPages/InstallOptionsWindow.axaml
GabrielDuf: fixed in the upcomming commit

MaxBuisson (approved):

aryanchoudharypro:
@GabrielDuf I've just found an issue with Avalonia, at least on windows, the framework doesn't send the proper textbox state to the screenreader, which meens that the text in the edit box will not be editable in an accessible way like deleteing or inserting characters, as it doesn't even tell you which character the nvda cursor is on, or what got deleted, ETC.

PR #4628, opened 2026-04-22
Revert "Harden Linux manager executable lookup to prevent PATH hijacking"

GabrielDuf:
Reverts Devolutions/UniGetUI#4621


mmorrissette-devolutions (approved):
LGTM

PR #4633, opened 2026-04-23
Enrich WinGet package names from COM API for bundled activation

GabrielDuf:
This pull request improves the handling of truncated package names when using the bundled in-process COM fallback for WinGet update detection. It introduces logic to enrich and restore full package names and IDs when they are truncated (ending with '…'), by cross-referencing results from the COM API.

**Enhancements to update detection and package name handling:**

* When falling back to the CLI for update detection in `GetAvailableUpdates_UnSafe`, the returned package list is now post-processed to restore truncated names using the new `EnrichTruncatedNamesFromCom` method.
* Added the `EnrichTruncatedNamesFromCom` method, which detects truncated package names or IDs (ending with '…') in CLI results and attempts to restore their full values by matching them against results from the COM API. If enrichment fails, a warning is logged and the original truncated names are retained.


GabrielDuf:
has been fix by https://github.com/Devolutions/UniGetUI/pull/4605

PR #4675, opened 2026-05-03
Make the main window resizeable on Linux

ekse:
The main window was not resizeable because the X11 Avalonia backend treats WindowDecorations.BorderOnly as WindowDecorations.None. This change implements resize grips on the window.

https://github.com/user-attachments/assets/f6f90d2f-db2c-474f-a696-f96c5dfbfb62

### DIFF
--- a/src/UniGetUI.Avalonia/Views/MainWindow.axaml.cs
+++ b/src/UniGetUI.Avalonia/Views/MainWindow.axaml.cs
@@ -5,6 +5,7 @@ using Avalonia.Controls;
 using Avalonia.Controls.ApplicationLifetimes;
 using Avalonia.Input;
 using Avalonia.Interactivity;
+using Avalonia.Layout;
 using Avalonia.Media;
 using Avalonia.Threading;
 using Avalonia.VisualTree;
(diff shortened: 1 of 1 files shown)


mamoreau-devolutions:
I'd just want to make sure that this doesn't affect WSL usage - there's a runtime check for WSL to adapt it and avoid some issues: https://github.com/Devolutions/UniGetUI/pull/4580

For regular Linux we can leave native decorations enabled but for WSLg it wouldn't get sized and maximized properly. Disabling decorations to get them from the Windows size instead worked better

Runtime checks for WSL are fine to adapt between the two

ekse:
> I'd just want to make sure that this doesn't affect WSL usage - there's a runtime check for WSL to adapt it and avoid some issues: #4580
> 
> For regular Linux we can leave native decorations enabled but for WSLg it wouldn't get sized and maximized properly. Disabling decorations to get them from the Windows size instead worked better
> 
> Runtime checks for WSL are fine to adapt between the two

This is taken into account, the border grips are enabled only when `isWsl` is false. (see https://github.com/Devolutions/UniGetUI/pull/4675/changes#diff-133754dff4c1cb9804c5fabcde40f72eb6a6c36bc32e2e3bcd175e50f5f3a963R211-R214)

mamoreau-devolutions (approved):

PR #4685, opened 2026-05-04
Update Korean

VenusGirl:
Update Korean


mamoreau-devolutions:
for some reason those two strings didn't get correctly synchronized in the last translation update batch, I'll have to see about fixing them for all remaining languages

mamoreau-devolutions (approved):

PR #4715, opened 2026-05-07
show "UniGetUI" in menu bar instead of "Avalonia Application"

GabrielDuf:
This pull request makes a small update to the `App.axaml` file by adding a `Name` attribute to the application definition. This change helps identify the application instance in the XAML markup.

* Added `Name="UniGetUI"` attribute to the root `Application` element in `App.axaml`.


jePerreault02 (approved):
LGTM

PR #4721, opened 2026-05-08
Updated translation and corrected translation failures caused by source code strings

StarsShine11904:
Hello, I have updated the Traditional Chinese translation and fixed the translation failure issue caused by capitalization issues in the source code strings

For details, please view the change records


StarsShine11904 (commented):
"Package Managers" should be changed to "Package managers" to resolve the issue of the string not being translated
This is a problem caused by the case of letters
Image

GabrielDuf (approved):

PR #4752, opened 2026-05-12
Log UI variant and environment info on startup

GabrielDuf:
## Summary
  - Adds a `UI Framework:` line and an environment block (OS, arch, runtime, elevated, packaged, args) to the startup log on both UI variants.
  - The Avalonia path was previously logging neither the textart banner nor any version/build info — it now mirrors the WinUI banner so bug reports from both variants look the same.
  - Adds a `CoreTools.IsPackagedApp()` helper (P/Invoke to `GetCurrentPackageFullName`) so we can distinguish MSIX vs unpackaged installs in logs.

Motivation: incoming bug reports often don't say which build the user is running, and there's no way to tell from the log. With this change, the first ~15 lines of every log file answer "WinUI or Avalonia, MSIX or portable, x64 or arm64, elevated or not, what args".


forestmaxime (approved):
LGTM

PR #4891, opened 2026-06-05
add xget github acceleration

yangweijie:
- [x] I have read the [contributing guidelines](https://github.com/Devolutions/UniGetUI/blob/main/CONTRIBUTING.md#coding), and I agree with the [Code of Conduct](https://github.com/Devolutions/UniGetUI/blob/main/CODE_OF_CONDUCT.md)**.
-----

add github softawre download acceleration  by https://github.com/xixu-me/Xget.git
-----

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/Generated Files/Secrets.Generated.cs
+++ b/Generated Files/Secrets.Generated.cs
@@ -1,0 +1,11 @@
+// Auto-generated file - do not modify
+namespace UniGetUI.Services
+{
+    internal static partial class Secrets
+    {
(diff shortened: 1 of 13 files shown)


GabrielDuf:
Thanks for the contribution.

After review, we've decided not to move forward with this feature due to security considerations for third-party download acceleration services due to security and trust considerations.

We'll be closing this PR. Thanks again for the contribution and your effort.

PR #4964, opened 2026-06-18
Avalonia UI: refine navigation, search boxes, filter pane & operation banners

GabrielDuf:
Addresses users feedback on the Avalonia UI, bringing several interactions closer to the WinUI app and smoothing out the animations. 

### Navigation panel
- Replaced the single width sidebar with a WinUI `NavigationView`-style layout: a static icon rail plus a labeled flyout that **cross-fades** in over the content (GPU-composited, so the package list never reflows during the transition).
- Rail and flyout are one `SidebarView` bound to the same view-model (new `ShowLabels` property), so selection, badges and loading stay in sync.
- Fixed 64px rail so loading bars/badges can't inflate it; loading indicator is a small bar centered under the icon.
- Transparent light-dismiss, collapse-on-navigate, and a title-bar back button wired to the existing back-navigation.
- Thinner, more Fluent hamburger glyph.

### Search boxes (top bar + Discover mega query)              
  - WinUI-style focus treatment: a single **accent underline** across the whole box (field + button) instead of recoloring the entire border.   
  - The field keeps the container fill in every state (no black background swap on focus), so field and button stay one consistent color.              
  - **Hover** shows a subtle neutral underline; focus switches it to the accent color.                                       
  - Top box: subtler corner radius and a small left nudge to match WinUI placement.
  
  ### Filter pane                                               
  - Opening/closing now **slides** on a compositor `RenderTransform` (clipped by the panel) instead of width-tweening the column, which reflowed the package list every frame and felt laggy. The list resizes once up front so the pane never sits beside an empty gap.                                     
  - Toolbar buttons keep sliding via their `MinWidth`transition. 
  
  ### Operations                                                
  - Dropped the redundant in-app success banner — the operations panel already shows the completed status. OS toasts and accessibility announcements are unchanged; failures still surface a banner.

### DIFF
--- a/src/UniGetUI.Avalonia/Infrastructure/AvaloniaOperationRegistry.cs
+++ b/src/UniGetUI.Avalonia/Infrastructure/AvaloniaOperationRegistry.cs
@@ -235,14 +235,6 @@ public static class AvaloniaOperationRegistry
 
         if (OperatingSystem.IsMacOS() && MacOsNotificationBridge.ShowSuccess(op))
             return;
-
-        if (TryGetMainWindow() is not { } mainWindow)
-            return;
-
-        mainWindow.ShowRuntimeNotification(
-            title,
-            message,
-            UniGetUI.Avalonia.Views.MainWindow.RuntimeNotificationLevel.Success);
     }
 
     private static void ShowOperationFailureNotification(AbstractOperation op)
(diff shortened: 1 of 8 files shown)


mamoreau-devolutions (approved):

GabrielDuf:
Hi @TonyBlur, just letting you know that I've added the few recommendations you mentioned the other day. Thanks for the feedback!

TonyBlur:
I really appreciate your effort and thank you for being so open to my feedback. ❤️

PR #4967, opened 2026-06-19
Update lang_nl.json

CateyeNL:
Changed "Herladen" to "Vernieuwen".

-----

-----

Closes #XXXX

Relates to #XXXX


CateyeNL:
Changed "Herladen" to "Vernieuwen".

GabrielDuf (approved):

PR #5084, opened 2026-07-14
Update CustomMessages.iss

bovirus:
@GabrielDuf 

- [X] **I have read the [contributing guidelines](https://github.com/Devolutions/UniGetUI/blob/main/CONTRIBUTING.md#coding), and I agree with the [Code of Conduct](https://github.com/Devolutions/UniGetUI/blob/main/CODE_OF_CONDUCT.md)**.
- [X] **Have you checked that there aren't other open [pull requests](https://github.com/Devolutions/UniGetUI/pulls) for the same changes?**
- [X] **Have you confirmed that this issue is caused by UniGetUI itself, and not by the package manager or the package involved?**

- Fixed Italian strings (note inb a multiline string about MSgBox strings is sueful to use %n as newline to separate paragraph and is missn g "." at the end.
- Fixed English strings (Added "English." as starting identifier"
- Add some strings for Italian in [CustoMmessages] and [Setup] to fix issues for InnOS etup 5.x / 6.x (fixed in inno SSEtup 7.x).

### DIFF
--- a/InstallerExtras/CustomMessages.iss
+++ b/InstallerExtras/CustomMessages.iss
@@ -1,15 +1,19 @@
-﻿[CustomMessages]
+[Messages]
+
+Italian.SetupWindowTitle=Installazione di {#MyAppName} {#MyAppVersion}
+
+[CustomMessages]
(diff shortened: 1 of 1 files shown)


bovirus:
@GabrielDuf 

Could you please merge this PR??

GabrielDuf (approved):

PR #5235, opened 2026-08-02
Apply and centralize smooth scrolling throughout the UI

Meteoni:
This PR centralizes smooth scrolling across the app instead of applying it only to package lists.

It adds inertia for DataGrid and ScrollViewer, supports future-proof two-axis scrolling, respects reduced-motion, and adjusts decay up slightly to 0.15s from 0.12s for a better touchpad experience without making it too sluggish on mouse wheel. 

Package icon loading is now also improved with bounded preloading instead of relying on lazy-load only.


Meteoni:
Hi @mamoreau-devolutions, when you have a chance, could you please review this? All checks are passing. Thanks!

src/UniGetUI.Avalonia/ViewModels/SoftwarePages/PackagesPageViewModel.cs
```diff
@@ -473,9 +475,37 @@ private void Loader_FinishedLoading(object? sender, EventArgs e)
...
+            const int batchSize = 4;
+            for (int start = 0; start < wrappers.Length; start += batchSize)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                int count = Math.Min(batchSize, wrappers.Length - start);
+                var tasks = new Task[count];
+                for (int index = 0; index < count; index++)
+                    tasks[index] = wrappers[start + index].EnsureIconLoadedAsync();
+
+                await Task.WhenAll(tasks).WaitAsync(cancellationToken).ConfigureAwait(false);
```
GabrielDuf, line 503: this ConfigureAwait(false) makes batch 2+ of the preload continue on a threadpool thread, so EnsureIconLoadedAsync() ends up calling LoadIconAsync off the UI thread.                                                                                                                                                     
                                                                                                                                                                     
That's a problem because of the cache-hit branch in LoadIconAsync (PackageCollections.cs:271), which sets IconBitmap = cached directly with no dispatcher. That line isn't new, but it was only ever called from the UI thread before, so it was fine. The preload is the first off-thread caller.                                 
                                                                                                                                                                     
On a revisit the icon cache is warm, so visible rows hit that branch off-thread → the Image binding updates Source off-thread → InvalidOperationException. (Off-screen rows are fine, no Image exists yet, which is why it's easy to miss.)                                                                                   
                                                                                                                                                                     
Simplest fix is to marshal the cache branch like the download branch already does:                                                                                 
                                                                                                                                                                     
```
  if (cached is not null)                                                                                                                                            
      await Dispatcher.UIThread.InvokeAsync(() =>                                                                                                                    
      {                                                                                                                                                              
          if (!token.IsCancellationRequested) IconBitmap = cached;                                                                                                   
      });
```

GabrielDuf (commented):
I left a comment, could you take a look when you get a chance?

Meteoni:
> I left a comment, could you take a look when you get a chance?

Will do.

Meteoni:
It's now resolved.

GabrielDuf (approved):

PR #5325, opened 2026-08-26
Uzbek language added

FlutterGenerator:
-----

-----

Closes #XXXX

Relates to #XXXX


GabrielDuf:
Hi @FlutterGenerator,

The check failed with the following error:

```text
lang_uz.json] Placeholder mismatch for key: {0} Install options are currently locked because {0} follows the default install options.
Translation: {0} hozircha standart o‘rnatish sozlamalariga amal qilayotgani uchun o‘rnatish sozlamalari qulflangan.
missing {0} (expected 2, found 1)
```

Could you please take a look?

I also asked Copilot to review the changes, as I don't speak Uzbek. If any of Copilot's comments are incorrect, feel free to mention it in the conversation and resolve them.

GabrielDuf (approved):

Issue #5158, opened 2026-07-20
[BUG] Update Loop (Unkown)


ShezzyYuuki:
### Please confirm these before moving forward

- [x] I have searched for my issue and have not found a work-in-progress/duplicate/resolved issue.
- [x] I have tested that this issue has not been fixed in the latest [(beta or stable) release](https://github.com/Devolutions/UniGetUI/releases/).
- [x] I have checked the [FAQ](https://github.com/Devolutions/UniGetUI#frequently-asked-questions) section for solutions.
- [x] This issue is about a bug (if it is not, please use the [correct template](https://github.com/Devolutions/UniGetUI/issues/new/choose)).

### UniGetUI Version

2026.2.5

### Windows version, edition, and architecture

Windows 11 IoT 26100.8875 x 64

### Describe your issue

For the past few days I have been having issues with the program, and I have been "updating" over and over the same programs over and over again. I also tried uninstalling and installing the apps, and they still show the same error. If you need more for me to show, do tell me. And I do apologize if this was an... easy fix.

### UniGetUI Log

```text
[20/7/2026 17:22:03] Found 8 available updates from Winget
[20/7/2026 17:22:03] Could not resolve installer hosts for ShiftUp.NIKKE version Unknown: Failed during source operation 'manifest' for 'winget': Version 'Unknown' was not found for package 'ShiftUp.NIKKE' in source 'winget'.
[20/7/2026 17:22:04] Could not resolve installer hosts for Smilegate.Stove version Unknown: Failed during source operation 'manifest' for 'winget': Version 'Unknown' was not found for package 'Smilegate.Stove' in source 'winget'.
[20/7/2026 17:22:04] Could not resolve installer hosts for c0re100.qBittorrent-Enhanced-Edition version 5.2.3: Failed during source operation 'manifest' for 'winget': Version '5.2.3' was not found for package 'c0re100.qBittorrent-Enhanced-Edition' in source 'winget'.
[20/7/2026 17:22:21] Unknown mimetype image/jpeg for icon https://cdn.brandfetch.io/idQEkAXA-z/w/400/h/400/theme/dark/icon.jpeg?c=1bxid64Mup7aczewSAYMX&t=1769771355360, aborting download
[20/7/2026 17:22:28] Loaded operation parameters for package id=Microsoft.VCRedist.2013.x64 on manager Winget and operation Update: update --id "Microsoft.VCRedist.2013.x64" --exact --source winget --silent --include-unknown --architecture x64
[...]

Issue #5042, opened 2026-07-05
[BUG] 4 updates are missing


bege10:
### Please confirm these before moving forward

- [x] I have searched for my issue and have not found a work-in-progress/duplicate/resolved issue.
- [x] I have tested that this issue has not been fixed in the latest [(beta or stable) release](https://github.com/Devolutions/UniGetUI/releases/).
- [x] I have checked the [FAQ](https://github.com/Devolutions/UniGetUI#frequently-asked-questions) section for solutions.
- [x] This issue is about a bug (if it is not, please use the [correct template](https://github.com/Devolutions/UniGetUI/issues/new/choose)).

### UniGetUI Version

2026.2.2

### Windows version, edition, and architecture

Windows 11 Professional 64 bit 25H2

### Describe your issue

4 Updates provided in winget are missing in UniGetUI

### UniGetUI Log

```text
[05.07.2026 21:02:12]    __  __      _ ______     __  __  ______
                        / / / /___  (_) ____/__  / /_/ / / /  _/
                       / / / / __ \/ / / __/ _ \/ __/ / / // /
                      / /_/ / / / / / /_/ /  __/ /_/ /_/ // /
                      \____/_/ /_/_/\____/\___/\__/\____/___/
                          Welcome to UniGetUI Version 2026.2.2
[05.07.2026 21:02:12]   
[05.07.2026 21:02:12] Build 107
[05.07.2026 21:02:12] UI Framework: Avalonia
[05.07.2026 21:02:12] Data directory C:\Users\bege\AppData\Local\UniGetUI
[05.07.2026 21:02:12] OS: Microsoft Windows 10.0.26200
[05.07.2026 21:02:12] Process arch: X64 (OS: X64)
[05.07.2026 21:02:12] Runtime: .NET 10.0.9
[05.07.2026 21:02:12] Loaded language locale: de
[05.07.2026 21:02:12] Elevated: False
[05.07.2026 21:02:12] Packaged (MSIX): False
[05.07.2026 21:02:12] Args: (none)
[05.07.2026 21:02:13] Command cargo-binstall.exe was not found on the system
[05.07.2026 21:02:14] Unable to retrieve secret for resource 'UniGetUI/GitHubAccessToken': Element nicht gefunden.
                      
                      Anmeldeinformationen können nicht aus dem Tresor abgerufen werden.
[...]

Issue #1164, opened 2023-06-23
[FEATURE REQUEST] Visible download progress bar


uuc110:
## Before submitting the issue

- [✓] **I have checked that the requested feature is not a duplicate or, if it is, it needs to be reopened**
- [✓] **I have checked that I am running the latest version of WingetUI**
- [✓] **I have checked the [FAQ](https://github.com/marticliment/WingetUI#frequently-asked-questions) section**
- [✓] **I have read the [contributing guidelines](https://github.com/marticliment/WingetUI/blob/main/CONTRIBUTING.md#feature-requests) and I agree with the [Code of Conduct](https://github.com/marticliment/WingetUI/blob/main/CODE_OF_CONDUCT.md)**

## Is your feature request related to a problem? Please describe.

It's not my problem, I just feel that marquee style progress isn't that good?, its better to know the progress on installation. it will help in knowing how much installation is left, If someone is in hurry and want to check how much it is left. it will be useful.

A clear and concise description of what the problem is. Ex. I'm always frustrated when I don't know how much download is left.

## Describe the solution you'd like

Well there should be a progress bar maybe be default windows green colour progress bar or you can set up theme to choose from.

marticliment:
This migt not be doable due to the nature of the command-line parser, but I am going go give it a try.

uuc110:
> This migt not be doable due to the nature of the command-line parser, but I am going go give it a try. 

Oh I see, thanks for consideration and reply but really love this software 🫶

2tbc1887:
+1 for this
If not possible, maybe a download speed indicator to show that progress happening?

BartmanAbyss:
+1

Bobrokus:
+1

marticliment:
Hello,

Latest WingetUI Beta does include thisn feature for Package Managers that do support visual progressbars:
![image](https://github.com/marticliment/WingetUI/assets/53119851/48ce5fa8-d8fd-4e21-9c1d-b213774da764)

PR #5390, opened 2026-09-17
Add determinate operation progress and ACKit repository workflow

Cynrath:
-----

### Summary

Progress feature:

- Generic operation progress pipeline added (OperationProgress model, throttled ProgressChanged event, formatter).
- WinGet Native COM structured progress wired in (InstallPackageAsync / UpgradePackageAsync / UninstallPackageAsync).
- UI supports determinate progress (IsIndeterminate=false, 0-100 value, stage + downloaded/total bytes).
- Operation-card mapping extracted to OperationCardProgressState so the ViewModel contract is unit-tested without Avalonia.
- Measured download throughput: optional `OperationProgress.BytesPerSecond`, calculated generically in `AbstractOperation` from cumulative byte samples (EMA alpha=0.3), so WinGet native COM, HTTP `DownloadOperation` and future byte-counter managers share one mechanism while mappers stay stateless. Formatter appends live `· 1,2 MB/s` (B/s–GB/s, `FormatAsSize` conventions, no ETA); first sample has no speed; stage/unknown/retry transitions reset.
- Fallback behavior preserved: no structured progress means the existing indeterminate animation.

ACKit repository integration (same PR):

- Real `ackit init` (ACKit 0.5.2): `ackit.yml`, canonical `AGENTS.md` workflow block, `CLAUDE.md`/`GEMINI.md`/copilot shims, 4 builtin skills.
- Extended root `AGENTS.md` (formatting discipline, Windows guards, localization, task/evidence, git hygiene, CI, completion criteria).
- 6 custom UniGetUI skills (`dotnet-build-test`, `avalonia-ui`, `package-manager-integration`, `winget-native`, `github-pr-ci`, `ackit-repo-workflow`); 16 skills total, 0 validation issues (strict translation-source-sync ref fixed).
- Policy pack, committed task/evidence workflow (`TASK-0001`), scan baseline, readiness gates, provider-aware context pack, `docs/ACKIT.md`, ACKit CI workflow.

### Problem

The operation progress bar only shows the indeterminate animation, so users never see real download progress even when the package manager (WinGet Native COM API) provides byte counts and percentages.

### Architecture

WinGet Native structured progress -> generic OperationProgress abstraction -> AbstractOperation.ProgressChanged -> OperationViewModel -> Avalonia ProgressBar.

- Generic model lives in UniGetUI.PackageEngine.Enums with no manager-specific types; UI never sees COM types.
- AbstractOperation.ReportProgress is thread-safe, deduplicated, and coalesced (sub-1pct deltas within 200ms); ResetProgress runs on every attempt so retries never show stale values.
- DownloadOperation reports HTTP byte progress; PackageOperation tries native WinGet COM first under #if WINDOWS, then falls back to CLI.
- OperationCardProgressState holds the exact card mapping (status plus progress to indeterminate/value/LiveLine); OperationViewModel only copies it onto bindable properties on the UI thread via Dispatcher.UIThread.Post.
- Existing MainWindow.axaml IsIndeterminate/Value/ItemStatus bindings already support determinate mode; no layout redesign.

### WinGet Native structured progress

- WinGetProgressMapper maps InstallProgress/UninstallProgress states: Queued/Finalizing stay indeterminate, Downloading prefers byte-based percent, Installing/Uninstalling only honor positive percentages, Finished maps to Completed.
- Percentage is clamped 0-100; NaN/Infinity map to unknown; BytesRequired==0 falls back to reported percent only when positive; BytesDownloaded greater than BytesRequired clamps without losing counters; no synthetic download-to-install total.
- CanUseNative rejects virtual sources, missing COM backend, non-WinGet managers, custom CLI args, and unelevated admin-required operations.
- TryApplyInstallOptions mirrors CLI scope/mode/hash/agreements/force/location/arch/version semantics; version pin resolves PackageVersionId from AvailableVersions or falls back to CLI.
- InterpretInstallResult mirrors CLI success paths and feeds MarkUpgradeAsDoneForNative so #5042 suppression and #5158 stuck-loop detection keep working (updates count, installs do not).

### Fallback semantics

- Managers without structured progress behave exactly as before (indeterminate).
- Unknown install progress falls back to indeterminate; queue/cancel/retry/cancellation semantics unchanged.
- Native pre-execution misses return null to CLI; mid-execution COM exceptions surface as Failure (no silent double-execution).
- Broker operations bypass native and keep existing broker behavior.
- New user-facing strings go through CoreTools.Translate with English fallback; AutomationProperties.ItemStatus binding untouched.

### Tests

Real commands run (Windows, .NET 10.0.401, x64, tr-TR system locale):

- Focused progress suites: `dotnet test src/UniGetUI.PackageEngine.Tests --filter OperationProgress|WinGetNativeProgress|OperationCardProgressState|DownloadOperationThroughput /p:Platform=x64`: 101/101 passed (portable TFM) + 111/111 passed (windows TFM), 0 failed. Covers first-sample-null, delta math, EMA determinism, zero-time/backward/repeated counters, stage/retry/unknown resets, non-Downloading sanitize, formatter units + NaN/Infinity omit, 8x50 parallel thread-safety, WinGet-mapped downstream speed, and a real throttled-loopback HTTP download test.
- Builds: `UniGetUI.PackageEngine.Operations` and `UniGetUI.Avalonia` (`/p:Platform=x64`): 0 errors (only pre-existing NU1903/NU1510/CA1822/CA2008/AVLN5001 warnings; CS2012 lock investigated and proven a transient stale-process lock via handle.exe + no UniGetUI/testhost processes + clean rebuild, no repo change needed).
- `dotnet format whitespace src --folder --verify-no-changes`: clean. `dotnet format style src/UniGetUI.Windows.slnx --no-restore --verify-no-changes`: clean (no bare mutating `dotnet format` used).
- Full PackageEngine suite: 1595 passed / 7 failed, and all 7 fail identically on the stashed baseline (no WIP changes): 5 are tr-TR locale assertions (`OperationHistory` marker 1 + `WinGetManagerTests` UpdateNotApplicable/HashMismatch message assertions 4) and 2 are launcher-asset discovery (`OperationCallArgsWiringTests` Scoop/PowerShell). No new failures from this change.

### Manual runtime verification

VERIFIED with a real local Debug x64 build (`UniGetUI.exe --headless`, `UNIGETUI_WINGET_COM=enabled`) driving Free Download Manager over the native WinGet COM path (op 8371548, 46.7 MB):

- `Starting native WinGet upgrade for SoftDeluxe.FreeDownloadManager...`
- Determinate percentage + downloaded/total, live: `Downloading · 2% · 1,0 MB / 46,7 MB` → … → `Downloading · 100% · 46,7 MB / 46,7 MB`
- Real measured throughput, varying live (no synthetic values; first sample correctly has no speed): `· 1,2 MB/s`, `· 985,8 KB/s`, `· 775,7 KB/s`, `· 1,3 MB/s`, …, `· 915,9 KB/s`
- Download→install transition drops the speed: `Installing...`, `Installing · 1%`, then `Finalizing...`, `Native WinGet result: Ok`, success message
- HTTP `package download` of the same installer also completed 0→100% with success and a 46.7 MB file on disk (its speed reaches the GUI card via the same `ProgressChanged` path; proven at unit level with a throttled-loopback test)

Observation (not a PR defect): the installed FDM version stayed 6.34.4.6974 because the vendor `.../6/latest/fdm_x64_setup.exe` payload itself reports FileVersion 6.34.4.6974 — stock `winget upgrade` CLI reproduces the identical "Successfully installed" with unchanged version. WinGet/UniGetUI reported the real `Ok` faithfully.

### Known limitations

- Install phase with no usable InstallationProgress stays indeterminate (Installing...); no synthetic total.
- Native path falls back to CLI for COM inactive, virtual sources, custom CLI args, version-lookup miss, unelevated admin-required ops, and broker ops.

### ACKit repository integration

- Actual ACKit init on this branch: `ackit.yml` (schemaVersion 1, `ackit config check` OK), `AGENTS.md` extended plus canonical managed block, `CLAUDE.md` (`@AGENTS.md`), `GEMINI.md` + copilot shims (all `ackit sync` up-to-date).
- Skills: 4 builtin + 6 custom UniGetUI + 6 existing translation = 16 total, `ackit skills validate` 0 issues. The historical strict `translation-source-sync` reference to `src/Languages/lang_en.json` is fixed (skill-relative `../../../src/...`), not pre-existing.
- Policy/config/task: `ackit policy check` OK, `ackit task doctor` OK (`TASK-0001` active in-PR with synchronized evidence; `docs/tasks` committed, not excluded).
- Readiness: 89/100 full (`--strict` pass); changed-scope 92. Baseline `docs/ackit/scan-baseline.json` commits the 71 pre-existing findings; full `ackit scan` shows 74 (the +3 are `ACKIT070` mutable-pin notices on the new workflow itself, kept repo-consistent instead of guessing SHAs).
- Context pack: `ackit pack --profile codex --max-tokens 50000` validated (20 instruction nodes, active task included).
- CI: `.github/workflows/ackit.yml` (least privilege, pinned `npx @cynrath/agent-context-kit@0.5.2`, gates config/policy/skills/task-doctor/readiness-strict/changed-scan plus baseline SARIF artifact, no duplicate .NET work). Docs: `docs/ACKIT.md`.
- Dogfooding notes: presence rules need `glob` (not `path`); `pattern` rules are forbidden-only; `policy check` chain stays 0 while rule packs evaluate during scan; copilot-shim shadowing and skill-link cycle diagnostics are expected delegation artifacts. No separate ACKit source changes were justified.

-----
Relates to #1164


GabrielDuf:
hi @Cynrath, After reviewing the PR in detail, I’m not going to merge it.

There are two major concerns with this PR.

### 1. ACKit does not belong in this PR

A significant part of this PR has nothing to do with operation progress. It integrates ACKit, a separate project maintained by you, into the UniGetUI repository.

This includes changes to the repository's CI, agent instructions, configuration, documentation, and development workflow, all specifically to integrate ACKit.

I do not want to introduce ACKit as a dependency of UniGetUI's CI or development workflow. This integration is also unrelated to the purpose of this PR and should not have been included.

### 2. The progress feature introduces major WinGet regressions

This PR is presented primarily as adding determinate operation progress, but it also changes how WinGet operations are executed.

For many operations, the established `winget.exe` execution path is replaced with the WinGet COM API. This bypasses behavior and workarounds that have accumulated in the existing implementation.

Among the issues identified during review:

* Existing `AutoRetry` paths are bypassed, including elevation, permissions, installer elevation restrictions, version fallback, and architecture/scope retry cases.
* `NoApplicableInstallers` sets the architecture/scope fallback state but does not actually perform the expected retry.
* Existing proxy handling is not carried over to the native path.
* `NoApplicableUpgrade` has different success/failure semantics from the existing implementation.
* Existing reboot-required result handling is not fully preserved.
* Existing failure interpretation, including not-applicable and installer-hash-mismatch handling, is bypassed.
* Operation history loses the normal WinGet return code and detailed CLI output, reducing the information available for troubleshooting.
* Native uninstall does not preserve all existing scope/version behavior.
* The native path relies on cached COM package information, which introduces additional state/staleness concerns that need to be accounted for.
* The new execution path does not have sufficient tests for the behavior it replaces. Most of the new tests cover progress plumbing rather than WinGet execution compatibility.
* Several new translatable strings are missing from the translation source.
* Progress throttling mixes the injected clock with `DateTime.UtcNow`, making that behavior inconsistent and not properly covered by the clock-controlled tests.
* The native logging path can still emit every COM progress callback, effectively bypassing some of the progress throttling.

These aren't minor review comments around the implementation of a progress bar. The PR changes a critical package-operation path and introduces a substantial regression risk for existing WinGet functionality.

Because of both the unrelated ACKit integration and the scope/number of regressions introduced by the WinGet execution changes, I don't think this PR is suitable to continue reviewing in its current form, so I'm closing it.

Issue #5031, opened 2026-07-01
[IMPROVEMENT] Improve Git upgrade failure handling when Git processes are running


SHJordan:
### Please confirm these before moving forward.

- [x] I have searched for my feature proposal and have not found a work-in-progress/duplicate/resolved/discarded issue.
- [x] This improvement refers to an existing feature. If you want to suggest a new feature, please use [this template](https://github.com/Devolutions/UniGetUI/issues/new?labels=new-feature&projects=&template=feature-request.yml&title=%5BFEATURE+REQUEST%5D+%28Enter+your+description+here%29).
- [x] This improvement is not a bug. If you want to report a bug, please use [this template](https://github.com/Devolutions/UniGetUI/issues/new?labels=bug&projects=&template=bug-issue.yml&title=%5BBUG%5D+%28Enter+your+description+here%29).

### Describe the improvement

Improve error reporting when a package installer fails because running processes are blocking the installation (e.g. Git).

## Current behavior

When updating Git through UniGetUI using WinGet, the installation may fail with a generic error.

In my case, UniGetUI reported only that the update had failed.

After inspecting the installer log generated by the package installer, the actual reason was:

```text
The following process(es) use Git for Windows:

Git for Windows (PID ...)
...

Please terminate those processes and retry.
```

The installer aborts because it is executed in silent mode (`/VERYSILENT`), so it cannot prompt the user to close the running processes.

## Expected behavior

**If possible**, it would be helpful if UniGetUI could surface the underlying installer error whenever it provides a more specific reason than the generic failure reported by WinGet.

For example, in this case, UniGetUI could display a message such as:

> Git is currently being used by one or more running processes. Please close the applications using Git (or terminate the listed processes) and try the update again.

Optionally, UniGetUI could offer a **Retry** button after the user has closed the blocking processes.

### Describe how this improvement could help users

Without checking the installer log, it is not obvious why the update failed. A clearer and more actionable error message would save users from having to manually inspect log files for a very common scenario.

This is not a WinGet or Git installer bug—only a suggestion to improve the user experience in UniGetUI by surfacing the actual installer failure reason whenever possible.

marco-carvalho:
Same UX class, with a case where WinGet already prints the reason on stdout (no installer-log parse needed).

Updating `Spotify.Spotify` while Spotify is running:

```text
The application is currently running. Exit the application and try again.
Installer failed with exit code: 26
```

[...]

PR #5398, opened 2026-09-18
Explain WinGet failures when the app is still running

### LINKED ISSUE #5397: [IMPROVEMENT] Detect installer "app is running" failures (Spotify exit 26) and offer close/retry

- [x] I have searched for my feature proposal and have not found a work-in-progress/duplicate/resolved/discarded issue.
- [x] This improvement refers to an existing feature. If you want to suggest a new feature, please use [this template](https://github.com/Devolutions/UniGetUI/issues/new?labels=new-feature&projects=&template=feature-request.yml&title=%5BFEATURE+REQUEST%5D+%28Enter+your+description+here%29).
- [x] This improvement is not a bug. If you want to report a bug, please use [this template](https://github.com/Devolutions/UniGetUI/issues/new?labels=bug&projects=&template=bug-issue.yml&title=%5BBUG%5D+%28Enter+your+description+here%29).

Related, but not the same:
- #5031 asks to surface Git installer-log text when processes block the update.
- #997 (closed as not planned in 2023) was Git blocked by running processes; at the time UniGetUI did not yet have `KillProcessOperation`.

This issue is about a case WinGet already explains on stdout, which UniGetUI currently maps to a generic failure.

### Describe the improvement

When updating `Spotify.Spotify` while Spotify is running, WinGet already prints a clear reason:

```text
Found Spotify [Spotify.Spotify] Version 1.3.1.234.g59d6bf59
Successfully verified installer hash
Starting package install...
The application is currently running. Exit the application and try again.
Installer failed with exit code: 26
```

Closing `Spotify.exe` and retrying the same `winget upgrade --id Spotify.Spotify` succeeds. In UniGetUI the same update only shows a generic failure (`OperationFailedDialog` title = `Metadata.FailureMessage`, plus “see the Command-line Output”).

WinGetPkgOperationHelper already maps several WinGet codes to `OperationVeredict.AutoRetry` (elevation, already installed, not applicable). Installer exit **26** / stdout containing **“currently running”** falls through to `OperationVeredict.Failure`.

UniGetUI already has the pieces to do better:
- `InstallOptions.KillBeforeOperation` + `KillProcessOperation`
- `RetryMode.Retry` / `Retry_Interactive` on `OperationFailedDialog`
- `Settings.K.KillProcessesThatRefuseToDie`

Suggested change (small, existing patterns):

1. In `WinGetPkgOperationHelper._getOperationResult`, detect this class of failure from stdout (`currently running` / localized equivalents) or installer exit `26` when WinGet reports `Installer failed with exit code: 26`.
2. Set a specific `FailureMessage`, e.g. `{package} is currently running. Close it and retry the update.`
3. On the failed-operation dialog, add **Retry after closing {process}** that prepends `KillProcessOperation` (package exe / `KillBeforeOperation`) and retries, same idea as retry-as-admin.
4. Optional: if `KillBeforeOperation` is empty, guess `Spotify` from `Spotify.Spotify` / display name, only when the installer said the app is running.

Do **not** silently `taskkill /F` by default. Reuse `CloseMainWindow()` first, then the existing “kill processes that refuse to die” setting.

`winget` CLI was run unelevated; elevation is not the issue here (Spotify also prohibits elevation).

### Describe how this improvement could help users

Spotify, Discord, Git, browsers, etc. often fail silent WinGet upgrades only because the app is open. UniGetUI currently looks like a hard error. Showing the real reason and offering close+retry would match what the CLI already says, and would reuse `KillProcessOperation` instead of asking users to discover “Close apps before this operation” in package options.

marco-carvalho:
-----

When a WinGet installer fails because the app is still running (for example Spotify with installer exit 26), UniGetUI currently shows only a generic error even though the CLI output already explains the failure.

This change:
- Detects that in-use condition from WinGet output (English/Portuguese markers, or installer exit code 26 when the log reports it)
- Replaces the generic failure message with a specific explanation
- Offers a retry that closes the guessed process first (`Retry after closing the running app`), with a flyout option to retry without closing
- Does **not** kill the app on the first attempt

Verified locally: 8 new WinGet unit tests plus 12 related existing verdict tests passed.

-----
Closes #5397
Relates to #5031


src/UniGetUI.PackageEngine.Operations/PackageOperations.cs
```diff
@@ -940,16 +945,86 @@ List<string> Output
...
+                if (
+                    !Options.KillBeforeOperation.Exists(existing =>
+                        existing.Equals(processName, StringComparison.OrdinalIgnoreCase)
+                    )
+                )
+                    Options.KillBeforeOperation.Add(processName);
+
+                AddPreOperation(
+                    new InnerOperation(
+                        new KillProcessOperation(processName, forceKill: true),
```
GabrielDuf, line 1073: `forceKill: true` bypasses the user's `KillProcessesThatRefuseToDie` setting, so the process is force-killed — `Kill(entireProcessTree: true)`, up to a second after `CloseMainWindow()` — regardless of what the user chose.

Worth noting that setting is off by default: boolean settings are file-presence based (SettingsEngine.cs:21), so this force-kills for everyone who hasn't explicitly turned it on, not just for people who opted out. The button says "closing", and the PR description doesn't mention the override, so nothing surfaces it.

Is that deliberate? The argument for it is reasonable — the user did click a button that says close the app — it just needs saying, and then the label should say force-close.

src/UniGetUI.PackageEngine.Operations/PackageOperations.cs
```diff
@@ -940,16 +945,86 @@ List<string> Output
...
+        {
+            var names = new List<string>();
+            AddProcessName(names, package.Name);
+            int separator = package.Id.LastIndexOf('.');
+            string idTail = separator >= 0 ? package.Id[(separator + 1)..] : package.Id;
+            AddProcessName(names, idTail);
+            return names;
+        }
+
+        private static void AddProcessName(List<string> names, string? candidate)
```
GabrielDuf, line 991: This can resolve to UniGetUI itself. `AssemblyName` is `UniGetUI` (UniGetUI.Avalonia.csproj:20), so on a failed WinGet update of the UniGetUI package, `package.Name` → `Process.GetProcessesByName("UniGetUI")` matches the running app — and the default button then       
  force-kills its own process tree, including the winget/pinget child running the operation and the elevator.                             
                                                                                                                                          
  A self-update failing because the app is running is exactly the case this feature targets, so I don't think it's hypothetical. Could `AddProcessName` skip the current process (`Process.GetCurrentProcess().ProcessName`)?

src/UniGetUI.PackageEngine.Operations/PackageOperations.cs
```diff
@@ -940,16 +945,86 @@ List<string> Output
...
+                    )
+                )
+                    Options.KillBeforeOperation.Add(processName);
+
+                AddPreOperation(
+                    new InnerOperation(
+                        new KillProcessOperation(processName, forceKill: true),
+                        mustSucceed: false
+                    )
+                );
```
GabrielDuf, line 1076: Under the agent broker this kills twice, and the local kill runs before policy is evaluated.

`_getPreInstallOps` deliberately returns before adding kill ops for broker-eligible packages (line 1190), with the comment that kill/pre/post are carried in the broker request "so that policy is evaluated before anything runs, and must not also be executed locally". This block does both: `Options.KillBeforeOperation.Add` (picked up by `BrokerRequestBuilder.Build` at line 373, which runs per attempt) and `AddPreOperation`, which `_runOperation` executes unconditionally.

Brokered operations do reach this code — the broker path calls `GetProcessVeredict` at line 769, so `FailedBecauseApplicationRunning` can be set there and the retry offered. Net effect: the app is killed locally before the broker sees (and possibly denies) the request, then killed again broker-side.

To be fair on severity: `UseAgentBroker` is opt-in and off by default, so this doesn't affect a default install today. It does quietly break the invariant that comment documents, which is the kind of thing that only surfaces once the broker is switched on by default.

Branching on `IsBrokerEligible(Package)` would fix it: local → `AddPreOperation` only, brokered → `Options.KillBeforeOperation` only. That also drops the `Options` mutation on the local path, which is dead there since pre-ops are snapshotted in the constructor.

src/UniGetUI.Avalonia/Views/DialogPages/OperationFailedDialog.axaml.cs
```diff
@@ -86,65 +86,115 @@ protected override void OnOpened(EventArgs e)
...
+
+        if (PackageOperation.CanRetrySkippingIntegrityChecks(
+                pkgOp.Package.Manager, pkgOp.Options, pkgOp.Role, pkgOp.WillRunElevated))
+            retryOptions.Add(MenuItem(CoreTools.Translate("Retry skipping integrity checks"),
+                () => { operation.Retry(AbstractOperation.RetryMode.Retry_SkipIntegrity); Close(); }));
+
+        if (!PackageOperation.CanRetryClosingRunningApp(pkgOp))
+            return;
+
+        retryOptions.Add(MenuItem(CoreTools.Translate("Retry without closing"),
```
GabrielDuf: After the first close-and-retry, this item stops doing what it says. `_closeRunningAppRetryQueued` (PackageOperations.cs:991) makes the queuing one-shot, but nothing removes the `KillProcessOperation` from `PreOperations` — so if the retry fails again and the user picks "Retry without closing" here, the app is still closed.

Either make the kill conditional on the retry mode at run time, or hide this item once a kill has been queued.

src/UniGetUI.PackageEngine.Managers.WinGet/Helpers/WinGetPkgOperationHelper.cs
```diff
@@ -433,6 +433,39 @@ private static void RecordUpgradeAttempt(string id, string version)
     internal static bool ReportedInstallerHashMismatch(int returnCode) =>
         (uint)returnCode is 0x8A150011;
 
+    internal static bool ReportedApplicationCurrentlyRunning(
```
GabrielDuf: WinGet documents locale-independent codes for this condition, and I'd make them the primary signal:

- `0x8A150101` `APPINSTALLER_CLI_ERROR_INSTALL_PACKAGE_IN_USE`
- `0x8A150103` `APPINSTALLER_CLI_ERROR_INSTALL_FILE_IN_USE`
- `0x8A150111` `APPINSTALLER_CLI_ERROR_INSTALL_PACKAGE_IN_USE_BY_APPLICATION`

(Not `0x8A150102`, "another installation is already in progress" — that needs a different remedy.)

The code is already available in the case this is built from. "The application is currently running. Exit the application and try again." is not installer output — it's winget's own `InstallFlowReturnCodePackageInUse` resource, and winget prints it only when the manifest maps the installer's code to `packageInUse`. That same branch sets `terminationHR = APPINSTALLER_CLI_ERROR_INSTALL_PACKAGE_IN_USE` (InstallFlow.cpp) and `winget.exe` exits with `context.GetTerminationHR()` (Command.cpp). So whenever that line appears, the return code is `0x8A150101` — one check, in all 11 languages winget ships, instead of two working string markers.

It would also match the convention in this file: `ReportedInstallerHashMismatch` (line 433) is a pure code check, and `ReportedUpdateNotApplicable` (line 469) only string-matches when `SelectedCliToolKind is BundledPinget` — i.e. only against output UniGetUI controls and knows is English.

src/UniGetUI.PackageEngine.Managers.WinGet/Helpers/WinGetPkgOperationHelper.cs
```diff
@@ -433,6 +433,39 @@ private static void RecordUpgradeAttempt(string id, string version)
...
+    [
+        "currently running",
+        "already running",
+        "em execução",
+        "en ejecución",
+        "in uso",
+        "package in use",
+        "being used by another process",
+        "processes use",
+    ];
```
GabrielDuf: Two of these don't match what winget actually prints. The line comes from winget's `InstallFlowReturnCodePackageInUse` resource, so here is every shipped translation:

| locale | winget's text |
| --- | --- |
| en-US | Application is **currently running**. Exit the application then try again. |
| pt-BR | O aplicativo está **em execução** no momento. Saia do aplicativo e tente novamente. |
| es-ES | La aplicación **se está ejecutando actualmente**. Salga de la aplicación e inténtelo de nuevo. |
| it-IT | L'applicazione è attualmente **in esecuzione**. Uscire dall'applicazione, quindi riprovare. |
| de-DE | Die Anwendung wird zurzeit ausgeführt. Beenden Sie die Anwendung, und versuchen Sie es noch mal. |
| fr-FR | L'application est en cours d'exécution. Quittez l'application, puis réessayez. |
| ja-JP | アプリケーションは現在実行中です。アプリケーションを終了してから、もう一度お試しください。 |
| ko-KR | 응용 프로그램이 현재 실행 중입니다. 애플리케이션을 종료한 후 다시 시도하세요. |
| ru-RU | Приложение сейчас запущено. Выйдите из приложения и повторите попытку. |
| zh-CN | 应用程序当前正在运行。退出应用程序，然后重试。 |
| zh-TW | 應用程式目前正在執行。結束應用程式，然後再試一次。 |

`currently running` and `em execução` hit. `en ejecución` does not — es-ES says "se está ejecutando actualmente". `in uso` does not either — it-IT says "in esecuzione". So the Spanish and Italian entries never fire, and `in uso` is loose enough to match unrelated text elsewhere.

Two more on the list:

- `being used by another process` is the generic Win32 sharing-violation wording — it can come from a locked log or temp file that has nothing to do with the app being open.
- The scan at line 441 has no return-code gate, and `output` isn't only winget's: `AbstractProcessOperation` collects every Error/Information line, which includes UniGetUI's own pre-operation banners and any `PreInstallCommand` output. A pre-install script echoing "already running" would trip it.

(Minor: the pt-BR test input, "Este aplicativo está em execução.", isn't winget's string either — it passes on the substring, so the test wouldn't catch a drift in the real one.)

All of which argues for the HRESULT check in the other comment: one condition instead of eleven strings.

src/UniGetUI.PackageEngine.Managers.WinGet/Helpers/WinGetPkgOperationHelper.cs
```diff
@@ -433,6 +433,39 @@ private static void RecordUpgradeAttempt(string id, string version)
...
         (uint)returnCode is 0x8A150011;
 
+    internal static bool ReportedApplicationCurrentlyRunning(
+        IReadOnlyList<string> processOutput,
+        int returnCode
+    )
+    {
+        if (processOutput.Any(LineReportsApplicationInUse))
+            return true;
+        if (returnCode is not 26)
```
GabrielDuf: This can only fire for the bundled pinget CLI — system winget never exits with a raw installer code. `winget.exe` returns `context.GetTerminationHR()` (Command.cpp), which `ReportInstallerResult` sets to either the manifest's mapped `ExpectedReturnCodes` HRESULT or the per-installer-type failure HR (`APPINSTALLER_CLI_ERROR_SHELLEXEC_INSTALL_FAILED` for an exe). `Installer failed with exit code: 26` is a line winget prints, not what winget exits with.

pinget does exit with it — `std::process::exit(result.exit_code)` — and prints `Failed to install X (exit code: 26)`, so both halves of this condition hold there.

So this branch is the pinget path while the markers above are the system-winget path. If that split is deliberate it's worth a comment saying so, because it isn't visible from the code. Either way, 26 only tells you an installer exited 26, not why.

src/UniGetUI.PackageEngine.Operations/PackageOperations.cs
```diff
@@ -940,16 +945,86 @@ List<string> Output
...
             }
 
             return Task.FromResult(veredict);
         }
 
+        public static bool CanRetryClosingRunningApp(PackageOperation operation)
+        {
+            if (!operation.FailedBecauseApplicationRunning)
+                return false;
+            return GuessCloseProcessNames(operation.Package).Count > 0;
```
GabrielDuf: `Count > 0` only says a candidate string survived the filter, not that a matching process exists. `Microsoft.VisualStudioCode` → `VisualStudioCode` (the real process is `Code`), `Python.Python.3.12` → `12` via the tail at line 971, and the space filter at line 982 drops every multi-word display name. In those cases the user is offered "Retry after closing the running app", it kills nothing, and the retry fails identically.

Cheapest improvement: only return true when `Process.GetProcessesByName` actually finds something. Sturdier: match processes whose `MainModule` path is under the package's ARP `InstallLocation`.

src/UniGetUI.PackageEngine.Operations/AbstractOperation.cs
```diff
@@ -50,7 +50,7 @@ public void ApplyCapabilities(bool admin, bool interactive, bool skiphash, strin
         BadgesChanged?.Invoke(this, new BadgeCollection(admin, interactive, skiphash, scope));
     }
 
-    private readonly IReadOnlyList<InnerOperation> PreOperations = [];
+    private readonly List<InnerOperation> PreOperations = [];
```
GabrielDuf, line 53: `PreOperations` was never mutated after construction before this; it's now appended from `ApplyRetryAction` (PackageOperations.cs:204). `_runOperation` only sets `Status = Running` *after* the pre-op `foreach` (line 460), and `Retry()` applies immediately when Status is `InQueue` — so an `AddPreOperation` landing during the pre-op phase mutates the list mid-enumeration.

I first wrote this off as unreachable, but I don't think it is. `ShowDetails` (OperationViewModel.cs:330) has no single-instance guard and uses `Show`, not `ShowDialog`, so a failed operation can have two dialogs open at once. Retry in the first one with any mode, then, while its pre-operations are running (`KillProcessOperation` alone waits up to a second per process), retry in the second one with "Retry after closing the running app": `_closeRunningAppRetryQueued` is still false, so it appends while the loop is enumerating. `MainThreadCore` catches it, so the user gets "An internal error occurred" and a failed operation rather than a crash.

Narrow, and arguably the pre-existing double-dialog is the real bug. But snapshotting the list before the loop makes it unreachable by construction again, which is where it was.

src/UniGetUI.Avalonia/Views/DialogPages/OperationFailedDialog.axaml.cs
```diff
@@ -86,65 +86,115 @@ protected override void OnOpened(EventArgs e)
...
-            if (OperatingSystem.IsWindows() && !pkgOp.Options.RunAsAdministrator && caps.CanRunAsAdmin)
-                retryOptions.Add(MenuItem(CoreTools.Translate("Retry as administrator"),
-                    () => { operation.Retry(AbstractOperation.RetryMode.Retry_AsAdmin); Close(); }));
+        return SplitRetryButton(operation, retryOptions, defaultRetryMode, defaultRetryLabel);
+    }
 
-            if (!pkgOp.Options.InteractiveInstallation && caps.CanRunInteractively)
-                retryOptions.Add(MenuItem(CoreTools.Translate("Retry interactively"),
-                    () => { operation.Retry(AbstractOperation.RetryMode.Retry_Interactive); Close(); }));
+    private static string DefaultRetryMode(AbstractOperation operation)
```
GabrielDuf: This makes the destructive action the primary button, with the safe one in the flyout. Given the PR description notes it doesn't kill on the first attempt, is inverting the default deliberate? Keeping "Retry" primary and putting the close-and-retry in the flyout would preserve that.

Separately, the label at line 109 sits in a `*`-width column of a two-column grid (OperationFailedDialog.axaml:36-41), and `Styles.Common.axaml:291` applies `TextTrimming="CharacterEllipsis"` to every `TextBlock`, which is what a `Button`'s string content renders into — so in longer translations this will ellipsize rather than wrap. Something like "Close app and retry" would survive.

src/UniGetUI.PackageEngine.Tests/WinGetManagerTests.cs
```diff
@@ -2228,6 +2228,159 @@ public async Task WinGetFailureUnrelatedToTheInstallerHashKeepsTheDefaultMessage
...
+            .WithManager(new WinGet())
+            .WithId("Spotify.Spotify")
+            .WithName("Spotify")
+            .Build();
+
+        Assert.Equal(["Spotify"], PackageOperation.GuessCloseProcessNames(package));
+    }
+
+    [Fact]
+    public void WinGetGuessesCloseProcessNamesFromTheIdTailWhenTheDisplayNameIsNotAProcess()
```
GabrielDuf: Nit: the name says the display name isn't a process, but the first case (`Git.Git` / "Git") does resolve from the display name — only the `SomeVendor.ToolName` case exercises the tail.

GabrielDuf (changes requested):
Thanks for the pr, I left a few inline notes, if you could take a look.

src/UniGetUI.PackageEngine.Operations/PackageOperations.cs
marco-carvalho: Yes, deliberate for this explicit action: the user asked to close the app so the retry can succeed, so we still force-kill after the 1s `CloseMainWindow()` wait.

The label is now `Force-close app and retry`, and that option is in the flyout rather than the primary button, so the override is visible.

src/UniGetUI.PackageEngine.Operations/PackageOperations.cs
marco-carvalho: Fixed. Close-and-retry now branches on `IsBrokerEligible(Package)`: local path uses per-attempt pre-ops only, brokered path adds to `Options.KillBeforeOperation` only. Those broker entries are cleared on any other retry mode, so policy still sees the kill before anything runs locally.

src/UniGetUI.PackageEngine.Operations/PackageOperations.cs
marco-carvalho: Done. `AddProcessName` now skips `Process.GetCurrentProcess().ProcessName`, so a failed UniGetUI self-update cannot queue a kill of this process tree.

src/UniGetUI.PackageEngine.Managers.WinGet/Helpers/WinGetPkgOperationHelper.cs
marco-carvalho: Done. Detection is now only those three HRESULTs (`0x8A150101`, `0x8A150103`, `0x8A150111`). `0x8A150102` is excluded. Locale string matching is gone.

src/UniGetUI.Avalonia/Views/DialogPages/OperationFailedDialog.axaml.cs
marco-carvalho: Removed that menu item. Kill is now per attempt: `Retry` is the primary action and does not queue a close, and `Force-close app and retry` only applies to that run. A later plain retry no longer inherits a leftover kill op.

src/UniGetUI.PackageEngine.Managers.WinGet/Helpers/WinGetPkgOperationHelper.cs
marco-carvalho: Agreed. The marker list is gone; this path is HRESULT-only now, so the Spanish/Italian misses, the sharing-violation wording, and ungated output scans are all out.

src/UniGetUI.PackageEngine.Managers.WinGet/Helpers/WinGetPkgOperationHelper.cs
marco-carvalho: Dropped the exit-code-26 path. It only told us the installer exited 26, not why, so it is no longer treated as in-use. Pinget and system WinGet now share the HRESULT check.

src/UniGetUI.PackageEngine.Operations/PackageOperations.cs
marco-carvalho: Done the cheap check: `CanRetryClosingRunningApp` is true only when `GetProcessesByName` finds a live process (after skipping this process). ARP `InstallLocation` matching is a better follow-up, not in this change.

src/UniGetUI.Avalonia/Views/DialogPages/OperationFailedDialog.axaml.cs
marco-carvalho: Inverted. Primary is `Retry` (no close). Force-close is in the flyout as `Force-close app and retry`, which should also survive the ellipsis style better.

src/UniGetUI.PackageEngine.Tests/WinGetManagerTests.cs
marco-carvalho: Renamed to `WinGetGuessesCloseProcessNamesFromDisplayNameAndIdTail`.

src/UniGetUI.PackageEngine.Operations/AbstractOperation.cs
marco-carvalho: Fixed. `_runOperation` now snapshots `PreOperations` plus per-attempt ops before the loop, and close-and-retry no longer appends to `PreOperations`. A second dialog cannot mutate the list mid-enumeration.

marco-carvalho:
Thanks for the review, @GabrielDuf. Addressed in dcc17a303:

- In-use detection is HRESULT-only (`0x8A150101`, `0x8A150103`, `0x8A150111`)
- Primary retry is the safe one; `Force-close app and retry` is in the flyout and is per-attempt
- Skip killing UniGetUI itself; only offer close when a matching process is actually running
- Broker vs local kill is split so policy still runs first
- Pre-ops are snapshotted so a second failure dialog cannot mutate the list mid-run

Ready for another look.

GabrielDuf (approved):
Thanks for taking the time to make these corrections. It looks good to me now. I'll merge it once the CI checks finish.

PR #5400, opened 2026-09-18
Fine tune UI 

Meteoni:
1. Scrolling: 
- Separate fractional-delta input from ordinary mouse-wheel input for better touchpad support.
- Add touchpad edge overpan with progressive resistance and spring return

2. Sidebar sliding pill: 
- Polish the pill's animation without introducing too much GPU cost. 

3. Settings:
- Fine-tune horizontal page-slide easing curve. 
- Keep accent border focus borders keyboard-only instead of also applying to mouse interactions

4. Shared theme resources
- Fix some of the badges having poor contrast due to using theme-aware accent colors. 
- Adjust dark-mode border alpha according to previous UI appearance. 

5. Filters pane: 
- Introduce dedicated border resource for accessibility

6. Installation options, Package Details & Screenshots:
- Replace the horizontal scrollbar with left / right chevrons
- Backdrop color adjustments
- Adaptive screenshot grids instead of fixed height
- Add direct screenshot touchpad dragging
- Use modern card styling instead of tabs which doesn't connect well with the content grid

Previous WinUI v.s. Current Avalonia post-adjustment comparison: 
- Avalonia: 
image
image
- Previous: 
image
image


src/UniGetUI.Avalonia/Views/SidebarView.axaml.cs
```diff
@@ -252,31 +262,50 @@ private async Task AnimatePillAsync(
...
 
         if (version != _pillAnimationVersion || cancellationToken.IsCancellationRequested)
             return;
 
-        _pillScale.ScaleY = 1d;
-        _pillTranslate.Y = targetCenter - (PillHeight / 2d);
+        SetPillEdges(targetTop, targetBottom);
+    }
+
+    private Task<TimeSpan> NextAnimationFrameAsync()
```
GabrielDuf: **Busy-spins the UI thread when the sidebar has no `TopLevel`.**

The null-`TopLevel` branch returns `Task.FromResult(TimeSpan.Zero)`. An already-completed task does not yield, so the `await NextAnimationFrameAsync()` inside `AnimatePillEdgesAsync` continues synchronously and the `while (true)` loop becomes a tight spin. Loop progress is driven by `Stopwatch`, not by frames, so it keeps spinning for the full `PillAnimationDuration` (300 ms) rather than exiting early.

The cancellation check at the top of the loop cannot break out either: `MoveSelectionPill` / `Cancel()` run on the same UI thread that is now blocked.

Repro: the pill animation is in flight (or gets started) while the sidebar is detached from its visual root — e.g. during main-window teardown.

Fix: `return` from the animation when there is no `TopLevel`, or await a real delay, instead of handing back a completed task.

src/UniGetUI.Avalonia/Views/Controls/SmoothScrollManager.cs
```diff
@@ -57,34 +89,191 @@ private static void OnTopLevelWheel(TopLevel topLevel, PointerWheelEventArgs e)
...
         }
-        e.Handled = true;
+
+        // A precision stream can occasionally land exactly on an integer. Retain the device
+        // classification briefly so one such sample does not switch to the mouse inertia path.
+        if (state.LastPrecisionTimestamp == 0 ||
+            Stopwatch.GetElapsedTime(state.LastPrecisionTimestamp, now).TotalSeconds > PrecisionGestureRetention)
+            return false;
+
+        state.LastPrecisionTimestamp = now;
```
GabrielDuf: **The retention window is refreshed by non-fractional deltas, latching mouse wheel onto the precision path.**

This line re-stamps `LastPrecisionTimestamp = now` on the retention fallback, i.e. for a delta that is *not* fractional. Every integral wheel notch that lands inside the `PrecisionGestureRetention` (120 ms) window therefore extends the window by another 120 ms.

Consequence: scroll the package list with the touchpad, then immediately spin the mouse wheel. Consecutive notches are comfortably under 120 ms apart, so the whole burst is classified as precision input for as long as the user keeps scrolling — inertia is disabled and each notch moves `PrecisionTouchpadDistance` (24 px) instead of `WheelDistance` (48 px).

The retention branch should classify as precision without re-stamping the timestamp; only a genuinely fractional delta (line 257) should extend the window.

Compounding this: `PackageDetailsWindow.OnScreenshotPointerWheelChanged` calls this same method a second time for the same event, re-stamping the shared per-`TopLevel` state.

src/UniGetUI.Avalonia/Views/DialogPages/PackageDetailsWindow.axaml.cs
```diff
@@ -121,32 +165,508 @@ private void OnVmPropertyChanged(object? sender, PropertyChangedEventArgs e)
...
+        if (_vm.ScreenshotCount == 0) return 0;
+        double width = GetScreenshotPageWidth();
+        double position = _screenshotSnapTarget ?? _screenshotPosition;
+        return Math.Clamp(
+            (int)Math.Round(position / width, MidpointRounding.AwayFromZero),
+            0,
+            _vm.ScreenshotCount - 1);
+    }
+
+    private void OnScreenshotPointerWheelChanged(object? sender, PointerWheelEventArgs e)
```
GabrielDuf: **This handler ignores `MotionPreference.ReducedMotion` but still swallows the event.**

`OnScreenshotPointerWheelChanged` unconditionally sets `e.Handled = true` for precision wheel events over `ScreenshotsBorder`, with no reduce-motion check. `OnTopLevelWheel` bails out on its first line under reduce-motion, so this tunnel handler is the only one left standing — and it then drives the animated free-pan through `ApplyScreenshotPanInput` (which, unlike `SnapScreenshotToIndex`, has no reduce-motion branch) and routes vertical deltas into `SmoothScrollManager.RoutePrecisionInput`.

Net effect for a user with *reduce motion* enabled: touchpad-scrolling over the screenshot area suppresses plain native scrolling and replaces it with exactly the custom motion the preference is supposed to turn off.

Fix: return early (leaving `e.Handled` false) when `MotionPreference.ReducedMotion` is set, or give `ApplyScreenshotPanInput` the same non-animated path `SnapScreenshotToIndex` has.

src/UniGetUI.Avalonia/Infrastructure/DirectionalSlideTransition.cs
```diff
@@ -60,9 +68,6 @@ public async Task Start(Visual? from, Visual? to, bool forward, CancellationToke
                 sv.VerticalScrollBarVisibility = ScrollBarVisibility.Auto;
         }
 
-        if (cancellationToken.IsCancellationRequested)
-            return;
-
         // Hide before clearing the transform so the outgoing page never snaps back on-screen.
```
GabrielDuf: **Dropping the post-await cancellation guard lets a cancelled transition clobber the next one.**

The `if (cancellationToken.IsCancellationRequested) return;` that used to sit between the `finally` and this cleanup was removed. `Animation.RunAsync` *completes* on cancellation rather than throwing, so a cancelled transition now falls straight through and runs `from.IsVisible = false; from.RenderTransform = null; to.RenderTransform = null;`.

Repro: in `SettingsBasePage` (`Frame.Content = page`), navigate A -> B and then B -> C before the 300 ms slide finishes. `TransitioningContentControl` cancels transition 1, whose `to` is page B — which is transition 2's `from`, currently mid-slide. Transition 1's stale continuation nulls B's `RenderTransform` (where Avalonia installed the slide's `TranslateTransform`), so the outgoing page visibly jumps mid-animation.

Restore the guard before the cleanup.

src/UniGetUI.Avalonia/Views/DialogPages/PackageDetailsWindow.axaml.cs
```diff
@@ -121,32 +165,508 @@ private void OnVmPropertyChanged(object? sender, PropertyChangedEventArgs e)
     {
         if (e.PropertyName == nameof(PackageDetailsViewModel.SelectedScreenshotIndex)
             || e.PropertyName == nameof(PackageDetailsViewModel.ScreenshotCount))
-            Dispatcher.UIThread.Post(UpdatePips);
+            Dispatcher.UIThread.Post(() =>
+            {
+                if (e.PropertyName == nameof(PackageDetailsViewModel.SelectedScreenshotIndex)
+                    && !_updatingScreenshotSelection
```
GabrielDuf: **This re-entrancy guard can never be observed as `true`.**

`SettleScreenshotSelection` sets `_updatingScreenshotSelection`, assigns `_vm.SelectedScreenshotIndex`, and clears it again — all synchronously. But `OnVmPropertyChanged` only *posts* the work to the dispatcher, and the posted lambda reads the flag long after it has been reset to `false`.

So `SetScreenshotPositionToIndex` always runs on the self-inflicted change. It is harmless today only by coincidence (`_screenshotSnapTarget is null` and the position already equals `index * pageWidth`), which makes it a latent trap for anyone touching this code later.

Either capture the flag at post time (`bool updating = _updatingScreenshotSelection;`) / hoist the check out of the lambda, or delete the guard.

src/UniGetUI.Avalonia/Views/Controls/SmoothScrollManager.cs
```diff
@@ -179,12 +454,186 @@ private bool ScrollBy(Vector step)
...
+        }
+
+        if (visual is null) return false;
+
+        _overpanVisual = visual;
+        _originalOverpanTransform = visual.RenderTransform;
+        _overpanTranslation = new TranslateTransform();
+        if (_target is DataGrid)
+        {
+            _originalRowsClipToBounds = visual.ClipToBounds;
```
GabrielDuf: **`ClipToBounds` can be permanently lost, plus two related gaps in the DataGrid overpan path.**

1. `visual.ClipToBounds = true` is set unconditionally here, but `ResetOverpan` only restores `_originalRowsClipToBounds` when `_overpanRows.Count > 0`. If `PART_RowsPresenter` has no realized visual children at attach time, the original value is never restored — and the next attach re-saves `true` as the "original", losing it for good.
2. This branch returns `true` without ever assigning `_appliedOverpanTransform`, so `ResetOverpan`'s `ReferenceEquals(_overpanVisual.RenderTransform, _appliedOverpanTransform)` check degenerates to `null == null` and will happily clear a `RenderTransform` that something else installed in the meantime.
3. Rows realized *after* attach never receive the shared `_overpanTranslation`, so they render un-offset while their siblings are overpanned.

src/UniGetUI.Avalonia/Views/Controls/SmoothScrollManager.cs
```diff
@@ -57,34 +89,191 @@ private static void OnTopLevelWheel(TopLevel topLevel, PointerWheelEventArgs e)
...
+            // synthesizing another fling here makes even a deliberate finger stop drift afterward.
+            _velocity = default;
+            _lastFrame = null;
+            Vector step = delta * SmoothScrollPhysics.PrecisionTouchpadDistance;
+            if (_target is DataGrid)
+            {
+                // DataGrid.UpdateScroll assigns (rather than adds to) its pending layout scroll.
+                // Multiple touchpad packets before one layout otherwise overwrite each other.
+                // Preserve every fractional delta and submit one combined step per frame.
+                if (_pendingPrecisionStep.X * step.X < 0 || _pendingPrecisionStep.Y * step.Y < 0)
```
GabrielDuf: **A sign flip on one axis discards the pending step on both axes.**

`_pendingPrecisionStep = default` zeroes the whole vector when only one component reversed direction. During a diagonal touchpad pan whose horizontal component wobbles across zero between two packets in the same frame, the accumulated *vertical* delta is silently thrown away, dropping a frame of scrolling.

Zero each axis independently:

```csharp
if (_pendingPrecisionStep.X * step.X < 0)
    _pendingPrecisionStep = _pendingPrecisionStep.WithX(0);
if (_pendingPrecisionStep.Y * step.Y < 0)
    _pendingPrecisionStep = _pendingPrecisionStep.WithY(0);
```

src/UniGetUI.Avalonia/Views/DialogPages/PackageDetailsWindow.axaml
```diff
@@ -201,6 +210,7 @@
 
                     <!-- ── Action button (split: main + chevron dropdown) ── -->
                     <Grid x:Name="ActionsPanel"
+                          Width="250"
```
GabrielDuf: **Hard-coded width will ellipsize long localized action labels.**

The inner `ColumnDefinition` already sizes to content with `MaxWidth="260"`, so this panel used to grow with `MainActionLabel` up to 260 px. `Width="250"` pins it regardless of the label.

Combined with the app-wide `TextBlock` trimming style, longer translations (German, Finnish, "Install and save options"-style strings) will now ellipsize where they previously fit. `MaxWidth="250"`, or leaving the content sizing alone, keeps the layout stable without truncating.

src/UniGetUI.Avalonia/Views/DialogPages/InstallOptionsControl.axaml
```diff
@@ -64,29 +84,58 @@
...
+            <Setter Property="Background" Value="Transparent"/>
+            <Setter Property="BorderThickness" Value="0"/>
+            <Setter Property="CornerRadius" Value="4"/>
+            <Setter Property="HorizontalContentAlignment" Value="Center"/>
+            <Setter Property="VerticalContentAlignment" Value="Center"/>
+        </Style>
+        <Style Selector="Button.install-options-tab-scroll:pointerover /template/ ContentPresenter">
+            <Setter Property="Background" Value="{DynamicResource SubtleFillColorSecondaryBrush}"/>
+        </Style>
+        <Style Selector="Button.install-options-tab-scroll:disabled">
```
GabrielDuf: **This `:disabled` style is dead code, and the chevrons it styles are now always visible.**

Nothing ever sets `IsEnabled=false` on `PreviousTabButton` / `NextTabButton`, and there is no visibility logic for them either. Since `HorizontalScrollBarVisibility` changed from `Auto` to `Hidden` on line 56, the two 32 px chevrons are permanently displayed and permanently enabled — including when every tab header already fits, where clicking them is a silent no-op at either end.

Either wire `IsEnabled` / visibility to the scroll extent and offset (which makes this style live), or drop the style along with the always-on buttons.

src/UniGetUI.Avalonia/Views/DialogPages/PackageDetailsWindow.axaml.cs
```diff
@@ -156,29 +676,57 @@ private void ApplyLayoutForCurrentSize()
...
+                RightPanel.IsVisible = false;
+                MainGrid.ColumnDefinitions[1].Width = new GridLength(0);
+            }
         }
-        else
+
+        UpdateScreenshotHeight();
+    }
+
+    private void UpdateScreenshotHeight()
```
GabrielDuf: **Height tracks the VM index while the strip tracks the visual index.**

`UpdateScreenshotHeight` derives the aspect ratio from `_vm.Screenshots[_vm.SelectedScreenshotIndex]`, but the pips and the strip translation use `GetVisualScreenshotIndex()`. During and after a swipe the container keeps the *previous* screenshot's height until `SettleScreenshotSelection` fires.

With screenshots of differing aspect ratios the panel height snaps in one step once the slide has already finished, instead of tracking the gesture. Using the visual index here (and interpolating between neighbours) would keep it in sync.

src/UniGetUI.Avalonia/Views/DialogPages/PackageDetailsWindow.axaml.cs
```diff
@@ -121,32 +165,508 @@ private void OnVmPropertyChanged(object? sender, PropertyChangedEventArgs e)
...
+    {
+        double remaining = Math.Max(0, ScreenshotEdgeOverpan - displacement);
+        if (remaining == 0 || input <= 0) return displacement;
+
+        double added = remaining *
+                       (1.0 - Math.Exp(-input * ScreenshotEdgeResistance / ScreenshotEdgeOverpan));
+        return displacement + added;
+    }
+
+    private async void CompleteScreenshotGesture(object? sender, EventArgs e)
```
GabrielDuf: **Needless `async void`.**

The body has no awaits — just a trailing `await Task.CompletedTask;`. The only thing `async` buys here is converting any exception thrown by `StartScreenshotSnapToRest` / `ResetScreenshotGestureRouting` from a normal handler exception into an async-void crash posted to the dispatcher.

Drop `async` and the trailing `await`.

src/UniGetUI.Avalonia/Infrastructure/DirectionalSlideTransition.cs
GabrielDuf, line 74: **Removing the post-await cancellation guard can leave the settings frame blank.**

This PR deletes the `if (cancellationToken.IsCancellationRequested) return;` that used to sit between the `finally` and this cleanup. That guard was load-bearing: Avalonia's `Animation.RunAsync` completes the task on cancellation (`cancellation = cancellationToken.Register(() => run.TrySetResult(null))`) rather than throwing, so a cancelled transition now falls straight through into the cleanup below.

Walking `TransitioningContentControl` with two presenters P1/P2 and `_isFirstFull` toggling per `UpdateContent`:

- Navigate A -> B: `from = P1(A)`, `to = P2(B)` — transition 1 starts.
- Navigate B -> C within the 300 ms: `UpdateContent` puts C in **P1** and sets `P1.IsVisible = true`; `ArrangeOverride` cancels transition 1 and starts transition 2 with `from = P2(B)`, `to = P1(C)`.
- Transition 1's stale continuation (queued on the dispatcher by the cancel) then runs its cleanup on *its* `from`/`to`: `P1.IsVisible = false`, `P1.RenderTransform = null`, `P2.RenderTransform = null`.

So it hides **P1 — the incoming page C** — and wipes the `TranslateTransform`s that transition 2 just installed on both presenters. When transition 2 finishes, `HideOldPresenter` hides P2 as well, and both presenters end up invisible: the content frame goes blank until the next navigation.

Reachable from `SettingsBasePage`, which sets `Frame.Content = page` on every navigation with this transition attached — two quick clicks through settings sub-pages is enough.

Please restore the guard before the cleanup block.

src/UniGetUI.Avalonia/Views/Controls/SmoothScrollManager.cs
GabrielDuf: **The retention window is refreshed by non-fractional deltas, latching mouse wheel onto the precision path.**

This line re-stamps `LastPrecisionTimestamp = now` in the retention fallback — i.e. for a delta that is *not* fractional. Every integral wheel notch landing inside the `PrecisionGestureRetention` window (0.12 s) therefore pushes the window forward another 120 ms, so the classification never decays as long as notches keep arriving.

Consequence: scroll the package list with the touchpad, then spin the mouse wheel. While notches stay under 120 ms apart (easy during a normal fast scroll) every one of them takes the precision path in `ApplyInput`, which does `_velocity = default; _lastFrame = null;` and applies `delta * PrecisionTouchpadDistance` directly — no inertia, and 24 px per notch instead of the 48 px `WheelDistance` the mouse path would give.

The retention branch should return `true` without re-stamping; only a genuinely fractional delta (line 257) should extend the window.

src/UniGetUI.Avalonia/Views/DialogPages/PackageDetailsWindow.axaml.cs
GabrielDuf, line 218: **This handler ignores `MotionPreference.ReducedMotion` but still swallows the event.**

`OnScreenshotPointerWheelChanged` is registered `RoutingStrategies.Tunnel` on `ScreenshotsBorder` and sets `e.Handled = true` for any precision wheel event, with no reduce-motion check (the only `ReducedMotion` reference in this file is in `SnapScreenshotToIndex`).

`SmoothScrollManager.OnTopLevelWheel` bails out on its first line under reduce-motion, so this tunnel handler is the only one left — and it drives:
- horizontal: `ApplyScreenshotPanInput`, which accumulates `_screenshotVelocity` and calls `RequestScreenshotFrame()`. No reduce-motion branch, unlike `SnapScreenshotToIndex`.
- vertical: `SmoothScrollManager.RoutePrecisionInput`, which goes straight to `RouteInput` and bypasses the `OnTopLevelWheel` guard entirely.

Net effect for a user with *reduce motion* on: touchpad-scrolling over the screenshot area suppresses plain native scrolling (tunnel + `Handled`) and replaces it with exactly the animated motion the preference is meant to switch off.

Fix: return early without setting `e.Handled` when `MotionPreference.ReducedMotion` is set.

src/UniGetUI.Avalonia/Views/SidebarView.axaml.cs
GabrielDuf: **A completed task here turns the pill animation into a UI-thread busy-spin.**

The null-`TopLevel` branch returns `Task.FromResult(TimeSpan.Zero)`. `await` on an already-completed task does not yield, so `AnimatePillEdgesAsync`'s `while (true)` keeps running synchronously on the UI thread. Loop progress comes from `Stopwatch`, not from frame callbacks, so it spins until wall-clock reaches `PillAnimationDuration` rather than exiting immediately. The `cancellationToken` check at the top of the loop cannot help either — `MoveSelectionPill` and `Cancel()` both run on the thread that is now blocked.

Scope, to be fair to the change: `MoveSelectionPill` only reaches the animation after `TranslatePoint` succeeds, so at *start* the sidebar is normally attached and this branch is cold. The realistic window is losing the `TopLevel` mid-animation (window teardown), and the spin is bounded at ~300 ms — so this is a latent hazard rather than something users hit daily.

Still worth fixing, since a "wait for the next frame" helper that can return synchronously is a trap for any future caller: `return` from the animation when there is no `TopLevel`, or await a real delay.

src/UniGetUI.Avalonia/Views/DialogPages/PackageDetailsWindow.axaml.cs
GabrielDuf: **This re-entrancy guard can never be observed as `true`.**

`SettleScreenshotSelection` sets `_updatingScreenshotSelection`, assigns `_vm.SelectedScreenshotIndex`, and clears it again, all synchronously. But `OnVmPropertyChanged` only *posts* the work (`DispatcherPriority.Loaded`), and the posted lambda reads the flag well after it has been reset to `false`.

So `SetScreenshotPositionToIndex` always runs on the self-inflicted change. It is harmless today only by coincidence — `_screenshotSnapTarget is null` and the position already equals `index * pageWidth` — which makes it a trap for whoever touches this next.

Either capture the flag at post time and check the captured value, or drop the guard and rely on the `_screenshotSnapTarget` / `_screenshotGestureAxis` conditions that are actually doing the work.

src/UniGetUI.Avalonia/Views/Controls/SmoothScrollManager.cs
GabrielDuf, line 555: **Asymmetric save/restore of `ClipToBounds`, plus two smaller gaps in the DataGrid overpan path.**

1. `_originalRowsClipToBounds` is saved and `ClipToBounds = true` applied unconditionally here, but `ResetOverpan` restores it only when `_overpanRows.Count > 0`. If `PART_RowsPresenter` has no realized visual children at attach time, `_overpanRows` stays empty, the restore is skipped, and the next attach re-saves the now-forced `true` as the "original" — the real value is gone for good. (Whether that is user-visible depends on what the DataGrid template sets by default; the asymmetry is worth closing regardless.)
2. This branch returns `true` without assigning `_appliedOverpanTransform`, so `ResetOverpan`'s `ReferenceEquals(_overpanVisual.RenderTransform, _appliedOverpanTransform)` becomes `ReferenceEquals(current, null)`. It only matches when the presenter's `RenderTransform` is null, and then assigns `_originalOverpanTransform` back onto it — i.e. it can re-install a stale transform on a presenter something else has since cleared. Harmless in the common case, but the guard isn't guarding anything.
3. Rows realized *after* attach never receive the shared `_overpanTranslation`, so they would render un-offset while their siblings are overpanned.

src/UniGetUI.Avalonia/Views/Controls/SmoothScrollManager.cs
GabrielDuf: **A sign flip on one axis discards the pending step on both axes.**

`_pendingPrecisionStep = default` zeroes the whole vector when only one component reversed. In a diagonal touchpad pan whose horizontal component wobbles across zero between two packets in the same frame, the accumulated *vertical* delta is thrown away too, dropping a frame of scrolling.

Zero each axis independently:

```csharp
if (_pendingPrecisionStep.X * step.X < 0)
    _pendingPrecisionStep = _pendingPrecisionStep.WithX(0);
if (_pendingPrecisionStep.Y * step.Y < 0)
    _pendingPrecisionStep = _pendingPrecisionStep.WithY(0);
```

src/UniGetUI.Avalonia/Views/DialogPages/InstallOptionsControl.axaml
GabrielDuf, line 105: **This `:disabled` style is dead code, and the chevrons it styles are always visible.**

Nothing ever sets `IsEnabled=false` on `PreviousTabButton` / `NextTabButton` — not in the template, not in `InstallOptionsControl.axaml.cs` — and there is no visibility logic either. Since this PR also swaps the tab-header `ScrollViewer` from `HorizontalScrollBarVisibility="Auto"` to `"Hidden"`, the two 32 px chevrons are now permanently shown and permanently enabled, including when every tab header already fits, where `ScrollTabHeaders` clamps to the same offset and the click is a silent no-op.

The WinUI `TabView` these mirror hide their scroll buttons when the headers fit. Binding `IsEnabled` (or `IsVisible`) to the scroll extent vs. viewport would make this style live and match that behaviour.

src/UniGetUI.Avalonia/Views/DialogPages/PackageDetailsWindow.axaml.cs
GabrielDuf, line 708: **Height tracks the VM index while the strip and pips track the visual index.**

`UpdateScreenshotHeight` derives the aspect ratio from `_vm.Screenshots[_vm.SelectedScreenshotIndex]`, whereas `UpdatePips` and the strip layout both use `GetVisualScreenshotIndex()`.

During a swipe the container therefore keeps the *previous* screenshot's height until `SettleScreenshotSelection` runs. With screenshots of differing aspect ratios the panel height changes in one step after the slide has already finished, instead of tracking it. Using the visual index here would keep the two in sync.

src/UniGetUI.Avalonia/Views/DialogPages/PackageDetailsWindow.axaml.cs
GabrielDuf: **`async` with nothing to await.**

The body has no awaits — just a trailing `await Task.CompletedTask;`. Marking the handler `async void` adds a state machine and moves any exception onto the dispatcher instead of the call site, for no benefit.

Drop `async` and the trailing `await`.

GabrielDuf (changes requested):
thanks for the pr, I've left a few comment for you to look at

Meteoni:
Thank you for the review. I've updated the pull request accordingly.

GabrielDuf (approved):
Looks good to me, thanks, I'll merge it

PR #5408, opened 2026-09-19
Output dialog family styling adjustments, splitter refinements, align package manager page controls, and other minor fixes

Meteoni:
- New restore down glyph
image

- Live output interface adjustments 
image

- Adjusted splitter appearance
image


src/UniGetUI.Avalonia/Views/MainWindow.axaml.cs
```diff
@@ -480,6 +495,13 @@ private void SetupResponsiveRail()
     public void RefreshNavigationMode()
         => ViewModel.Sidebar.Mode = SidebarViewModel.ParseMode(Settings.GetValue(Settings.K.NavMenuMode));
 
+    // Let the splitter hover surface replace the separator visually without changing its layout thickness.
+    private void OperationsSplitter_PointerEntered(object? sender, PointerEventArgs e)
+        => OperationsPanelBorder.Classes.Set("splitter-active", true);
+
+    private void OperationsSplitter_PointerExited(object? sender, PointerEventArgs e)
```
GabrielDuf: `splitter-active` can get stuck. It's set on `PointerEntered` and cleared only on `PointerExited`, but nothing clears it when the splitter disappears out from under the cursor - e.g. the last operation finishes and the panel auto-hides, or the panel collapses, while the pointer is still on the splitter. `OperationsPanelBorder` then keeps `BorderBrush="Transparent"` and the 1-DIP separator is missing the next time the panel shows.

`UpdateOperationsPanelRow()` already runs on every visibility/expansion change, so clearing the class there when `OperationsSplitterVisible` is false would cover it.

src/UniGetUI.Avalonia/Views/MainWindow.axaml
```diff
@@ -87,6 +87,40 @@
...
+        <!-- WinUI-like splitter: the track fills the resize gap while a compact pill
+             marks the drag axis. -->
+        <Style Selector="GridSplitter.winui-horizontal-splitter">
+            <Setter Property="Background" Value="Transparent"/>
+            <Setter Property="Template">
+                <ControlTemplate>
+                    <!-- Match the visible splitter track to the operation-card width.
+                         OperationsList uses 8 DIPs of horizontal inset. -->
+                    <Border Background="{TemplateBinding Background}"
+                            Margin="8,0"
```
GabrielDuf: This margin removes the drag affordance it's inset from. `GridSplitter` is a `TemplatedControl` that renders nothing itself - hit-testing follows the template - so with `Margin="8,0"` on the template root the outer 8 DIPs at each end of the splitter are no longer draggable. The stock template was full-bleed (`<Border Background="{TemplateBinding Background}"/>`).

`Padding="8,0"` on a full-bleed outer `Border`, with the rounded hover surface as an inner child, gets the same visual inset while keeping the whole track grabbable.

src/UniGetUI.Avalonia/Views/SoftwarePages/AbstractPackagesPage.axaml
```diff
@@ -98,6 +98,33 @@
...
+      <!-- The full splitter track is interactive. Its subtle rounded hover surface sits beneath
+           the persistent grip, matching WinUI's pane-resize affordance. -->
+      <Style Selector="GridSplitter.winui-vertical-splitter">
+         <Setter Property="Background" Value="Transparent"/>
+         <Setter Property="Template">
+            <ControlTemplate>
+               <!-- Match the visible splitter track to the package-list container on the
+                    right, whose outer card is inset 6 DIPs vertically. -->
+               <Border Background="{TemplateBinding Background}"
+                       Margin="0,6"
```
GabrielDuf: Same as the horizontal splitter in `MainWindow.axaml`: `Margin="0,6"` on the template root makes the top and bottom 6 DIPs of the filter-pane splitter undraggable, since the control has no background of its own to hit-test against.

src/UniGetUI.Avalonia/Views/DialogPages/OperationFailedDialog.axaml
```diff
@@ -4,36 +4,76 @@
...
+        <Style Selector="SplitButton.operation-dialog-action">
+            <Setter Property="Height" Value="32"/>
+            <Setter Property="MinWidth" Value="0"/>
+            <Setter Property="CornerRadius" Value="4"/>
+            <Setter Property="Background" Value="{DynamicResource SettingsCardBackground}"/>
+            <Setter Property="BorderBrush" Value="{DynamicResource SecondaryActionBorderBrush}"/>
+            <Setter Property="BorderThickness" Value="{DynamicResource SecondaryActionBorderThickness}"/>
+            <Setter Property="FontSize" Value="14"/>
+        </Style>
+        <Style Selector="SplitButton.operation-dialog-action:pointerover">
```
GabrielDuf: I don't think these two setters take effect. Avalonia's `SplitButton` paints through `PART_PrimaryButton` / `PART_SecondaryButton`, and those inner button themes set their own background from `SplitButtonBackground*` on hover/press - which beats a `Background` setter on the outer `SplitButton`.

So the retry split button here will most likely hover with the *global* `SplitButtonBackgroundPointerOver` this PR introduces (`ControlFillColorSecondaryBrush`), not `SettingsCardHoverBackground`. Worth hovering both halves to confirm; if it is inert, either drop these two styles or scope them to the template parts.

src/UniGetUI.Avalonia/Views/DialogPages/OperationFailedDialog.axaml
```diff
@@ -4,36 +4,76 @@
...
-        MaxHeight="550"
+        MaxHeight="490"
         MinHeight="300"
+        TitleMargin="24,0,0,0"
         Background="{DynamicResource AppDialogBackground}">
 
-    <Grid RowDefinitions="Auto,*,Auto">
+    <dialogs:ImmersiveDialog.Styles>
+        <!-- Scoped to this dialog family: these do not alter global Button resources. -->
+        <Style Selector="Button.operation-dialog-action">
```
GabrielDuf: This block is `secondary-action` plus `Height="32"` and `MinWidth="0"` - except it keeps the brushes (`SettingsCardBackground` / `SettingsCardHoverBackground` / `SubtleFillColorTertiaryBrush`) that this same PR migrates `secondary-action` *away* from. Net result: the same visual role hovers to opaque `#323232` in this dialog and to translucent `#15FFFFFF` everywhere else.

`Classes="secondary-action"` plus a local `Height`/`MinWidth` override would keep the two in sync. If the stronger hover is deliberate because the footer sits on the darker `AppWindowBackground`, the comment should say that instead.

Minor: the comment says "this dialog family", but the other two members (`OperationOutputWindow`, `OperationHistoryLogDialog`) don't use these styles - the history dialog uses `secondary-action`.

src/UniGetUI.Avalonia/Infrastructure/OperationHistoryLogDialog.cs
```diff
@@ -96,19 +99,41 @@ public static async Task ShowAsync(OperationHistoryRecord record)
...
+        dialog.Content = new Grid
+        {
+            RowDefinitions = new RowDefinitions("*,Auto"),
+            Children = { body, footerSurface },
         };
 
         await dialog.ShowDialog(owner);
     }
 
+    private static IBrush LookupBrush(string key, ThemeVariant theme, IBrush fallback)
```
GabrielDuf: Third copy of this helper - it's already in `ViewModels/DialogPages/OperationOutputViewModel.cs:64` and `Views/DialogPages/OperationFailedDialog.axaml.cs:76`, byte-for-byte. Good candidate to move onto `ThemeHelper` next to `Variant`.

Also, `Brushes.White` as the fallback for `TextFillColorPrimaryBrush` would be invisible in light theme if the lookup ever missed; the sibling in `OperationOutputViewModel` uses a theme-neutral literal.

src/UniGetUI.Avalonia/Views/SoftwarePages/AbstractPackagesPage.axaml
```diff
@@ -569,6 +597,7 @@
                                     <DataGridTemplateColumn.CellTemplate>
                                         <DataTemplate x:DataType="pkg:PackageWrapper">
                                             <CheckBox IsChecked="{Binding IsChecked}"
+                                                      Classes="package-selector package-list-selector"
```
GabrielDuf: `package-selector` is a dead class token - nothing anywhere selects `CheckBox.package-selector`; only `package-list-selector` has a style. Same on lines 765, 836 and 881, where `package-selector` is the only class added.

src/UniGetUI.Avalonia/Assets/Styles/Styles.Common.axaml
```diff
@@ -53,7 +54,29 @@
                         <GradientStop Offset="0" Color="#0F000000"/>
                         <GradientStop Offset="1" Color="#29000000"/>
                     </LinearGradientBrush>
+                    <SolidColorBrush x:Key="ControlFillColorDefaultBrush" Color="#B3FFFFFF"/>
+                    <SolidColorBrush x:Key="ControlFillColorSecondaryBrush" Color="#80F9F9F9"/>
+                    <SolidColorBrush x:Key="ControlFillColorTertiaryBrush" Color="#4DF9F9F9"/>
+                    <SolidColorBrush x:Key="ControlFillColorDisabledBrush" Color="#4DF9F9F9"/>
+                    <SolidColorBrush x:Key="ControlStrokeColorDefaultBrush" Color="#0F000000"/>
+                    <SolidColorBrush x:Key="ControlStrokeColorSecondaryBrush" Color="#29000000"/>
```
GabrielDuf: `ControlStrokeColorSecondaryBrush` is defined in both theme dictionaries and referenced zero times.

Related, just below: the new dark `ControlElevationBorderBrush` hardcodes `#18FFFFFF` / `#12FFFFFF` rather than `StaticResource`-ing the two stroke tokens defined immediately above it.

src/UniGetUI.Avalonia/Views/MainWindow.axaml.cs
GabrielDuf: `splitter-active` can get stuck. It's set on `PointerEntered` and cleared only on `PointerExited`, but nothing clears it when the splitter disappears out from under the cursor - e.g. the last operation finishes and the panel auto-hides, or the panel collapses, while the pointer is still on the splitter. `OperationsPanelBorder` then keeps `BorderBrush="Transparent"` and the 1-DIP separator is missing the next time the panel shows.

`UpdateOperationsPanelRow()` already runs on every visibility/expansion change, so clearing the class there when `OperationsSplitterVisible` is false would cover it.

src/UniGetUI.Avalonia/Views/MainWindow.axaml
GabrielDuf: This margin removes the drag affordance it's inset from. `GridSplitter` is a `TemplatedControl` that renders nothing itself - hit-testing follows the template - so with `Margin="8,0"` on the template root the outer 8 DIPs at each end of the splitter are no longer draggable. The stock template was full-bleed (`<Border Background="{TemplateBinding Background}"/>`).

`Padding="8,0"` on a full-bleed outer `Border`, with the rounded hover surface as an inner child, gets the same visual inset while keeping the whole track grabbable.

src/UniGetUI.Avalonia/Views/SoftwarePages/AbstractPackagesPage.axaml
GabrielDuf: Same as the horizontal splitter in `MainWindow.axaml`: `Margin="0,6"` on the template root makes the top and bottom 6 DIPs of the filter-pane splitter undraggable, since the control has no background of its own to hit-test against.

src/UniGetUI.Avalonia/Views/DialogPages/OperationFailedDialog.axaml
GabrielDuf: I don't think these two setters take effect. Avalonia's `SplitButton` paints through `PART_PrimaryButton` / `PART_SecondaryButton`, and those inner button themes set their own background from `SplitButtonBackground*` on hover/press - which beats a `Background` setter on the outer `SplitButton`.

So the retry split button here will most likely hover with the *global* `SplitButtonBackgroundPointerOver` this PR introduces (`ControlFillColorSecondaryBrush`), not `SettingsCardHoverBackground`. Worth hovering both halves to confirm; if it is inert, either drop these two styles or scope them to the template parts.

src/UniGetUI.Avalonia/Views/DialogPages/OperationFailedDialog.axaml
GabrielDuf: This block is `secondary-action` plus `Height="32"` and `MinWidth="0"` - except it keeps the brushes (`SettingsCardBackground` / `SettingsCardHoverBackground` / `SubtleFillColorTertiaryBrush`) that this same PR migrates `secondary-action` *away* from. Net result: the same visual role hovers to opaque `#323232` in this dialog and to translucent `#15FFFFFF` everywhere else.

`Classes="secondary-action"` plus a local `Height`/`MinWidth` override would keep the two in sync. If the stronger hover is deliberate because the footer sits on the darker `AppWindowBackground`, the comment should say that instead.

Minor: the comment says "this dialog family", but the other two members (`OperationOutputWindow`, `OperationHistoryLogDialog`) don't use these styles - the history dialog uses `secondary-action`.

src/UniGetUI.Avalonia/Infrastructure/OperationHistoryLogDialog.cs
GabrielDuf: Third copy of this helper - it's already in `ViewModels/DialogPages/OperationOutputViewModel.cs:64` and `Views/DialogPages/OperationFailedDialog.axaml.cs:76`, byte-for-byte. Good candidate to move onto `ThemeHelper` next to `Variant`.

src/UniGetUI.Avalonia/Views/SoftwarePages/AbstractPackagesPage.axaml
GabrielDuf: `package-selector` is a dead class token - nothing anywhere selects `CheckBox.package-selector`; only `package-list-selector` has a style. Same on lines 765, 836 and 881, where `package-selector` is the only class added.

src/UniGetUI.Avalonia/Assets/Styles/Styles.Common.axaml
GabrielDuf: `ControlStrokeColorSecondaryBrush` is defined in both theme dictionaries and referenced zero times.

Related, just below: the new dark `ControlElevationBorderBrush` hardcodes `#18FFFFFF` / `#12FFFFFF` rather than `StaticResource`-ing the two stroke tokens defined immediately above it.

src/UniGetUI.Avalonia/Assets/Styles/Styles.Common.axaml
```diff
@@ -238,8 +291,8 @@
                     <SolidColorBrush x:Key="SettingsCardBackground"      Color="#0DFFFFFF"/>
                     <SolidColorBrush x:Key="SettingsCardBorderBrush"     Color="#19000000"/>
                     <SolidColorBrush x:Key="SettingsCardHoverBackground" Color="#323232"/>
-                    <SolidColorBrush x:Key="SecondaryActionBorderBrush"  Color="Transparent"/>
-                    <Thickness x:Key="SecondaryActionBorderThickness">0</Thickness>
+                    <StaticResource x:Key="SecondaryActionBorderBrush" ResourceKey="ControlStrokeColorDefaultBrush"/>
+                    <Thickness x:Key="SecondaryActionBorderThickness">1</Thickness>
```
GabrielDuf: Worth flagging in the description too: dark `secondary-action` buttons and `ButtonCard`s now gain a 1-DIP border, and since `BorderThickness` participates in layout they also grow 2 DIPs in each axis. Fine in roomy settings pages, but worth a look at the tighter rows.

GabrielDuf (approved):
lgtm

Issue #5237, opened 2026-08-02
[BUG] Automatic restart after dependency installation crashes on macOS because powershell.exe is used


tromm:
### Please confirm these before moving forward

- [x] I searched for an existing issue and did not find a matching report.
- [x] I reproduced this with the latest installed release, UniGetUI 2026.2.7.
- [x] This report describes a bug.

### Environment

- UniGetUI: 2026.2.7, build 107
- macOS: 26.5.2 (25F84)
- Architecture: ARM64
- Hardware: MacBook Air M5 (Mac17,4)
- PowerShell: 7.6.4 installed as `/opt/homebrew/bin/pwsh`

### Description

After UniGetUI installs a missing package-manager dependency, it offers to restart itself. Selecting the restart option closes UniGetUI, but the application is not relaunched. On the next manual launch, UniGetUI reports the exception from the failed restart.

This is reproducible after installing dependencies such as `cargo-update`, `cargo-binstall`, or the Git dependency offered by the vcpkg integration.

### Steps to reproduce

1. Start UniGetUI on macOS.
2. Enable a manager with a missing dependency, for example Cargo.
3. Let UniGetUI install the dependency.
4. Accept the offered UniGetUI restart.

### Expected behavior

UniGetUI closes and relaunches on macOS.

### Actual behavior

UniGetUI closes, fails to relaunch, and records this exception:

```text
Exception type: Win32Exception (System.ComponentModel.Win32Exception)
Crash HResult: 0x80004005
Crash Message: An error occurred trying to start process 'powershell.exe'
with working directory '/'. No such file or directory

at System.Diagnostics.Process.Start(ProcessStartInfo)
at UniGetUI.Core.Tools.CoreTools.ScheduleRelaunchAfterExit(String)
at UniGetUI.Avalonia.Infrastructure.AppRestartHelper.Restart()
at UniGetUI.Avalonia.Views.DialogPages.MissingDependencyDialog.<OnInstallClickedAsync>d__7.MoveNext()
```

### Probable cause

In tag `v2026.2.7`, `CoreTools.ScheduleRelaunchAfterExit` unconditionally uses:

```csharp
FileName = "powershell.exe"
Arguments = $"-NoProfile -WindowStyle Hidden -Command \"{command}\""
```

Source:

https://github.com/Devolutions/UniGetUI/blob/v2026.2.7/src/UniGetUI.Core.Tools/Tools.cs

[...]

PR #5416, opened 2026-09-21
Detect git on macOS and Linux for the vcpkg dependency check

### LINKED ISSUE #5238: [BUG] vcpkg dependency check uses git.exe and Winget installation command on macOS

- [x] I searched for an existing issue and did not find a matching report.
- [x] I reproduced this with the latest installed release, UniGetUI 2026.2.7.
- [x] This report describes a bug.

### Environment

- UniGetUI: 2026.2.7, build 107
- macOS: 26.5.2 (25F84)
- Architecture: ARM64
- Hardware: MacBook Air M5 (Mac17,4)
- Git: `/usr/bin/git`, version `2.50.1 (Apple Git-155)`
- vcpkg executable: `/opt/homebrew/bin/vcpkg`
- `VCPKG_ROOT=/Users/tromm/vcpkg`

### Description

When vcpkg is enabled on macOS, UniGetUI reports that Git is missing even though Git is installed, on PATH, and works normally.

The missing-dependency dialog proposes this Windows-only command:

```text
winget install --id Git.Git --exact --source winget
```

Accepting the installation cannot provide a macOS `git.exe`. It also leads into the separate broken automatic-restart path that invokes `powershell.exe`.

### Steps to reproduce

1. Install vcpkg and Git on macOS.
2. Verify:
   ```sh
   command -v git
   git --version
   command -v vcpkg
   ```
3. Configure a valid `VCPKG_ROOT`.
4. Start UniGetUI and enable vcpkg.

### Expected behavior

UniGetUI detects `git` on non-Windows platforms and considers the dependency satisfied.

### Actual behavior

UniGetUI reports Git as missing and proposes installing Windows Git through Winget.

### Probable cause

In tag `v2026.2.7`, the vcpkg dependency is defined without an operating-system check:

```csharp
new ManagerDependency(
    "Git",
    CoreData.PowerShell5,
    "-ExecutionPolicy Bypass ... winget install --id Git.Git ...",
    "winget install --id Git.Git --exact --source winget",
    async () => (await CoreTools.WhichAsync("git.exe")).Item1
)
```

Source:

https://github.com/Devolutions/UniGetUI/blob/v2026.2.7/src/UniGetUI.PackageEngine.Managers.Vcpkg/Vcpkg.cs

On macOS the executable is `git`, not `git.exe`, regardless of whether Apple Git or Homebrew Git is installed.

### Workaround

Mark the Git dependency dialog as skipped. Git and vcpkg continue to work normally outside this incorrect dependency check.

awss1i:
-----
On macOS and Linux, enabling vcpkg makes UniGetUI report Git as missing on every start, even with `git` on PATH, and the dialog offers `winget install --id Git.Git`, which cannot run there. This PR makes the check look for the platform's `git`, gives the Install button a command that fits the platform, and makes the dialog say "terminal" instead of "Windows PowerShell prompt" off Windows.

### Why it happened

`Vcpkg.cs` declared the dependency with `WhichAsync("git.exe")`. `CoreTools.Which` appends PATHEXT only on Windows (`EnumerateCandidateFileNames` in `Tools.cs`); everywhere else the name is used literally, so `git.exe` can only match a file called `git.exe`. The same file already picks `vcpkg.exe` or `vcpkg` per platform for its own binary. Vcpkg is registered on every platform (`PEInterface.cs`), and `AvaloniaBootstrapper.GetMissingDependenciesAsync` checks every ready manager at startup, so Linux is affected exactly like the macOS report.

### What changed

**Commit 1: `Detect git on macOS and Linux for the vcpkg dependency check`**
- `VcpkgGitDependency.cs` (new) now builds the dependency; `Vcpkg.cs` just calls `VcpkgGitDependency.Create()`.
- Windows: same executable name (`git.exe`), same PowerShell arguments, same manual `winget` command. The strings were moved, not edited (compared against `main` byte for byte, see below).
- Off Windows: the check looks for `git`. The install command is chosen through `ManagerDependency`'s existing `resolveCommand` hook when the user clicks Install (the same mechanism Cargo uses), because the elevator is configured while the managers load:
  - macOS: `brew install git` when Homebrew is on PATH, or at `/opt/homebrew/bin/brew` / `/usr/local/bin/brew` (GUI-launched processes have a bare PATH); otherwise `xcode-select --install`.
  - Linux: the first of `apt-get`, `dnf`, `pacman`, `zypper` found on PATH, run through the configured elevator (`CoreData.ElevatorPath` + `ElevatorArgs`, so `sudo -A ...` or `pkexec ...`). The manual command shown in the dialog is the plain `sudo <manager> install ... git` form.
  - None found: the Install click raises a translated "No supported package manager was found. Please install Git manually and restart UniGetUI.", which the dialog already renders as "An error occurred: ...".
- `Resolve()` is a pure function taking the platform flag, a PATH lookup, a file-exists lookup and the elevator, so every outcome is unit-tested on any OS, including Windows CI.
- `UniGetUI.PackageEngine.Tests` gets a `ProjectReference` to the Vcpkg project and `InternalsVisibleTo`, like Homebrew and Cargo.

**Commit 2: `Say "in a terminal" off Windows in the missing-dependency dialog`**
- `MissingDependencyDialog` picks a new string off Windows; the existing Windows string is untouched, so no locale loses its translation. Each commit adds its one English key next to the existing dialog strings in `lang_en.json`.

### Verification

Fedora 44 host, .NET SDK 10.0.401, git 2.55.0, vcpkg 2026-07-27 bootstrapped with `VCPKG_ROOT` set. Debug build of `UniGetUI.Avalonia.slnx` run in portable mode with vcpkg enabled.

| Platform | Case | main | this branch |
|---|---|---|---|
| Linux | startup, git on PATH | `Command git.exe was not found on the system` / `Dependency Git not found for manager vcpkg.` / dialog shown | `Dependency Git for vcpkg is present.`, no dialog |
| Linux | startup, git hidden from PATH | dialog: "Windows PowerShell prompt", `winget install --id Git.Git --exact --source winget` | dialog: "in a terminal:", `sudo dnf install -y git` |
| Linux | what Install starts | `pwsh -ExecutionPolicy Bypass ... "& {winget install ...}"` (fails: no pwsh, and no winget either way) | the app had picked `sudo -A` with its askpass; `GetInstallCommand()` = `/usr/bin/sudo` `-A dnf install -y git`; started through a recording stand-in elevator it received exactly that argv |
| Linux | no supported package manager on PATH | not reachable | manual text is the "install manually" message; Install raises it, shown as "An error occurred: ..." |
| Windows | dependency strings | | identical to `main` (extracted and compared); `Create_OnWindows_KeepsTheWingetInstallCommand` pins them in CI |
| macOS | | | **not run**: I have no Mac. The three macOS outcomes are covered by the pure `Resolve()` tests only. @tromm, if you can start this build with vcpkg enabled, the dialog should no longer appear with Apple Git on PATH. |

- Tests: `UniGetUI.PackageEngine.Tests` 1232 → 1247 passed (15 new: 11 facts and a 4-row Theory), the 30 failures on Linux are the same pre-existing Windows-only ones before and after (`StartMenuShortcutsDatabaseTests`, `InstallOptionsFactoryTests`); `UniGetUI.Core.Tools.Tests` unchanged at 353 / 4. The 4 Theory rows and the 3 macOS facts run on Windows too.
- The detection test plants a fake `git` (spelled out in the test, not taken from the code under test) on the process PATH and asks the real `Vcpkg` manager. Setting `ExecutableName` back to `git.exe` on every platform (what `main` does) fails 2 tests on Linux; dropping `-y` from the dnf arguments fails the Theory row. On Windows the detection test passes before and after, since Windows was never affected.
- Gates on Linux: `dotnet format whitespace src --folder --verify-no-changes` and `dotnet format style UniGetUI.Avalonia.slnx --verify-no-changes` clean; `dotnet restore` + `dotnet test UniGetUI.Avalonia.slnx` from a fresh clone with an empty NuGet cache gives the same numbers. I could not run the PowerShell translation scripts locally (no pwsh); the JSON is valid with matching `{0}` placeholders and the `validate-translations` job covers it.
- Not clicked: the Install button inside the dialog. Synthetic input does not reach the window on my Wayland session, and a real click would prompt for my sudo password. The third row above is the substitute and uses the dialog's exact `ProcessStartInfo`.
- Cost: off Windows, `Create()` runs one to four PATH lookups at construction to fill the manual command. On this host that adds four log lines at startup, one at ImportantInfo (`Command apt-get was not found on the system`). If you would rather keep construction silent, the manual text can be made generic.

### Review guide

- `Vcpkg.cs`: 8 lines replaced by one call.
- `VcpkgGitDependency.cs`: the Windows branch of `Create()` is the old code verbatim; the rest is new. `Resolve()` is the only logic worth reading closely.
- `VcpkgGitDependencyTests.cs`: the PATH-replacing tests sit in a `DisableParallelization` collection.
- `MissingDependencyDialog.axaml.cs`: one ternary.

### Not in this PR

- #5237 (the restart after a dependency install spawns `powershell.exe` off Windows): separate code path, I can follow up with it.
- On macOS, `/usr/bin/git` is Apple's shim and exists before the Command Line Tools are installed; this PR treats it as present, the same as `command -v git` does.
- Other distributions' package managers (apk, xbps, emerge) are not resolved; the "install manually" message covers them.
-----
Closes #5238
Relates to #5237


GabrielDuf (approved):
lgtm

PR #5425, opened 2026-09-22
Operation Card Presentation Improvements, Settings Text Wrapping Fixes, and Other Misc Accessibility Improvements 

### LINKED ISSUE #5272: [IMPROVEMENT] New User Interface is worse than before

- [x] I have searched for my feature proposal and have not found a work-in-progress/duplicate/resolved/discarded issue.
- [x] This improvement refers to an existing feature. If you want to suggest a new feature, please use [this template](https://github.com/Devolutions/UniGetUI/issues/new?labels=new-feature&projects=&template=feature-request.yml&title=%5BFEATURE+REQUEST%5D+%28Enter+your+description+here%29).
- [x] This improvement is not a bug. If you want to report a bug, please use [this template](https://github.com/Devolutions/UniGetUI/issues/new?labels=bug&projects=&template=bug-issue.yml&title=%5BBUG%5D+%28Enter+your+description+here%29).

### Describe the improvement

Why do we have big red boxes now when package(s) fail(s) to update? The previous style was clean, sleek, with better colors. Please return to the original version instead of making it worse just for the sake of changing.

### Describe how this improvement could help users

Why do we have big red boxes now when package(s) fail(s) to update? The previous style was clean, sleek, with better colors. Please return to the original version instead of making it worse just for the sake of changing.

### LINKED ISSUE #5410: [BUG] Settings dialog is messed in Russian locale

- [x] I have searched for my issue and have not found a work-in-progress/duplicate/resolved issue.
- [x] I have tested that this issue has not been fixed in the latest [(beta or stable) release](https://github.com/Devolutions/UniGetUI/releases/).
- [x] I have checked the [FAQ](https://github.com/Devolutions/UniGetUI#frequently-asked-questions) section for solutions.
- [x] This issue is about a bug (if it is not, please use the [correct template](https://github.com/Devolutions/UniGetUI/issues/new/choose)).

### UniGetUI Version

2026.3.0

### Windows version, edition, and architecture

Microsoft Windows [Version 10.0.22631.7582]

### Describe your issue

Image

Image

One text overlays the other and controls.

### Steps to reproduce the issue

Set Russian locale and open package update settings.

### UniGetUI Log

```text
Not needed
```

### Package Managers Logs

```text
Not needed
```

### Relevant information

_No response_

### Screenshots and videos

_No response_

Meteoni:
- restore the compact operation-card layout, status styling, progress strip, badge flyouts, and finished-operation behavior. Closes #5272
image

- audit every element that might have border adornment against previous UI to ensure accessibility

- improve SettingsCard wrapping. Closes #5410

2026-09-22-011057-UniGetUI exe


src/UniGetUI.Avalonia/Views/MainWindow.axaml
```diff
@@ -301,144 +345,172 @@
...
+
+                                            <Image Grid.Column="0"
+                                                   Source="{Binding PackageIcon}"
+                                                   Width="32" Height="24"
+                                                   Margin="0"
+                                                   automation:AutomationProperties.AccessibilityView="Raw"
+                                                   VerticalAlignment="Center"
+                                                   Stretch="Uniform"/>
+
+                                            <TextBlock Grid.Column="1"
```
GabrielDuf: The title now sits in an `Auto` column but keeps `TextTrimming="CharacterEllipsis"`. An `Auto` column is sized to the text's full desired width, so the trimming can never engage.

With a long title — e.g. `KillProcessOperation`'s `"Closing process(es) {0}"` with several process names, or a long installer name — the `*` live-line column collapses to 0 and the grid's desired width exceeds the panel. The hosting `ScrollViewer` has `HorizontalScrollBarVisibility="Disabled"`, so the overflow is clipped rather than scrollable: the Cancel/Close button (col 4) and the `…` button (col 5) get pushed off-screen and become unreachable. Window `MinWidth` is 700, so this is hit at normal sizes.

Previously the title lived in a `*` column, where the ellipsis worked. Suggest giving the title the `*` column (and letting the live line take `Auto`/a capped width), or putting a `MaxWidth` on the title.

src/UniGetUI.Avalonia/Views/Controls/Settings/SettingsCard.cs
```diff
@@ -212,43 +199,87 @@ public SettingsCard()
...
         _border.Classes.Add("settings-card");
 
         base.Content = _border;
 
         PointerPressed += OnPointerPressed;
         KeyDown += OnKeyDown;
-        GotFocus += (_, e) =>
+        SizeChanged += (_, e) => UpdateResponsiveLayout(e.NewSize.Width);
+        // Keyboard focus is rendered by the app-wide Avalonia FocusAdorner. Do not also
+        // recolor/thicken the card border; the double outline is not WinUI-like.
```
GabrielDuf: The comment says keyboard focus is rendered by the app-wide Avalonia `FocusAdorner`, but no such adorner exists here. `grep -rn FocusAdorner src/UniGetUI.Avalonia` returns exactly one hit — `Styles.Common.axaml:815` — and that one *nulls* the adorner for `TextBox.search-field-standalone`.

`SettingsCard` derives from `UserControl`, which gets no `FocusAdorner` from Avalonia's Fluent theme (that property is set per-control in the Button/CheckBox/etc. control themes; its default is `null`). Meanwhile `IsClickEnabled` still sets `Focusable = true` and `OnKeyDown` still activates on Enter/Space, and the `Border.settings-card-keyboard-focused` style was deleted from `Styles.Common.axaml` — so a keyboard user can tab to and invoke a clickable settings card with no visual indication of where focus is.

If the goal was to drop the double outline, an explicit `FocusAdorner` on the card (or restoring a single focus visual) would still be needed.

src/UniGetUI.Avalonia/Views/Controls/Settings/SettingsCard.cs
```diff
@@ -212,43 +199,87 @@ public SettingsCard()
...
-            if (!_isClickEnabled || e.NavigationMethod == NavigationMethod.Pointer) return;
-            _border.Classes.Add("settings-card-keyboard-focused");
-            _border.BorderThickness = FocusedBorderThickness(_baseBorderThickness);
-        };
-        LostFocus += (_, _) =>
+            _iconPresenter.IsVisible = _headerIcon is not null;
+            return;
+        }
+
+        bool wrapContent = _rightContent is not null && width <= ContentWrapThreshold;
```
GabrielDuf: Two things on this breakpoint:

1. **No hysteresis.** The wrap state flips on a single threshold based on the card's own width. Settings pages host these cards in a `StackPanel` inside a `ScrollViewer` with auto scrollbars; wrapping makes every card on the page taller at once, which can bring the vertical scrollbar in, which shrinks available width by ~12px. If the pane width sits in that band around the threshold, you get unwrap → page shortens → scrollbar goes → width grows → wrap again, i.e. a self-sustaining layout loop. A small dead band between the wrap and unwrap thresholds would avoid it.
2. **The measured width isn't the content width.** This measures the outer `SettingsCard`, but `Border.settings-card` carries `Margin="40,0"`, so the effective content breakpoint is ~396, not the CommunityToolkit 476 the constant cites.

src/UniGetUI.Avalonia/Assets/Styles/Styles.Common.axaml
```diff
@@ -289,7 +317,7 @@
                     <StaticResource x:Key="AccentButtonForegroundDisabled" ResourceKey="TextOnAccentFillColorDisabledBrush"/>
                     <!-- Settings cards -->
                     <SolidColorBrush x:Key="SettingsCardBackground"      Color="#0DFFFFFF"/>
-                    <SolidColorBrush x:Key="SettingsCardBorderBrush"     Color="#19000000"/>
+                    <SolidColorBrush x:Key="SettingsCardBorderBrush"     Color="Transparent"/>
```
GabrielDuf: Dark-theme `SettingsCardBorderBrush` becomes `Transparent` here (and at `Styles.WindowsMica.axaml:65`), while the light theme keeps `#0F000000`.

Grouped settings cards rely on that brush for their shared divider — e.g. `General.axaml:36-48` and `Administrator.axaml:106-113` stack cards with `CornerRadius="8,8,0,0"` + `BorderThickness="1,0,1,1"`. In dark mode every one of those dividers now paints transparent, so a group of stacked cards renders as one undifferentiated `#0DFFFFFF` block with no separators — asymmetric with light mode, and the opposite of the accessibility goal of the PR.

src/UniGetUI.Avalonia/Assets/Styles/Styles.WindowsMica.axaml
```diff
@@ -17,21 +17,25 @@
...
             <!-- Cards / panels: translucent white layers over the (light) Mica -->
             <SolidColorBrush x:Key="SettingsCardBackground"      Color="#B3FFFFFF"/>
-            <SolidColorBrush x:Key="SettingsCardBorderBrush"     Color="#14000000"/>
+            <SolidColorBrush x:Key="SettingsCardBorderBrush"     Color="#0F000000"/>
             <!-- Package list: near-transparent so Mica shows through, matching WinUI's SystemFillColorNeutralBackground -->
             <SolidColorBrush x:Key="PackageListBackground"       Color="#06000000"/>
             <SolidColorBrush x:Key="SettingsCardHoverBackground" Color="#99FFFFFF"/>
             <SolidColorBrush x:Key="AppSplitterBackground"       Color="#18000000"/>
-            <SolidColorBrush x:Key="AppOperationsPanelBackground" Color="#80FFFFFF"/>
+            <SolidColorBrush x:Key="AppOperationsPanelBackground" Color="Transparent"/>
```
GabrielDuf: `AppOperationsPanelBackground` is set to `Transparent` in the light Mica dictionary, but the dark Mica dictionary (line 70) still has `#66202020`. With the new self-contained operation cards that gives a bare Mica backdrop behind the operations panel in light mode and a tinted panel in dark mode — one of the two was probably meant to change with the other.

src/UniGetUI.Avalonia/Infrastructure/AvaloniaOperationRegistry.cs
```diff
@@ -84,16 +84,15 @@ public static void Add(AbstractOperation op)
             Dispatcher.UIThread.Post(UpdateTrayStatus);
         };
 
-        // Cancellation drives Status = Canceled from several code paths, so StatusChanged(Canceled)
-        // can fire more than once for a single operation. Handle the terminal cancel exactly once.
+        // Cancellation can report Canceled from several code paths. Remove its progress
+        // notification once, but keep the finished card visible until the user closes it,
+        // matching the removed WinUI operation-control behavior.
```
GabrielDuf: Keeping the canceled card visible until the user closes it is clearly intentional, but dropping `RemoveAfterDelayAsync` from this path also drops the only reachable call to `IpcOperationApi.ForgetTracking(op.Metadata.Identifier)` — it was only invoked via `Remove` / `RemoveAfterDelayAsync`. A user who cancels operations and never clicks Close now leaves every canceled operation tracked in `IpcOperationApi` and present in `Operations` / `OperationViewModels` for the process lifetime.

Minor follow-on: with the delayed removal gone, the `Interlocked.Exchange(ref cancelHandled, …)` guard below protects only `RemoveProgress`, which is already idempotent, so the guard no longer buys anything.

src/UniGetUI.Avalonia/Views/MainWindow.axaml
GabrielDuf: The title now sits in an `Auto` column but keeps `TextTrimming="CharacterEllipsis"`. An `Auto` column is sized to the text's full desired width, so the trimming can never engage.

With a long title — e.g. `KillProcessOperation`'s `"Closing process(es) {0}"` with several process names, or a long installer name — the `*` live-line column collapses to 0 and the grid's desired width exceeds the panel. The hosting `ScrollViewer` has `HorizontalScrollBarVisibility="Disabled"`, so the overflow is clipped rather than scrollable: the Cancel/Close button (col 4) and the `…` button (col 5) get pushed off-screen and become unreachable. Window `MinWidth` is 700, so this is hit at normal sizes.

Previously the title lived in a `*` column, where the ellipsis worked. Suggest giving the title the `*` column (and letting the live line take `Auto`/a capped width), or putting a `MaxWidth` on the title.

src/UniGetUI.Avalonia/Views/Controls/Settings/SettingsCard.cs
GabrielDuf: Worth double-checking this one: `AdornerLayer.DefaultFocusAdorner` *is* set app-wide at `Styles.Common.axaml:406-418`, so a keyboard-focused card does still get a ring even with `settings-card-keyboard-focused` gone. The problem is *where* it lands.

The adorner adorns **this control** — the `SettingsCard` wrapper — not the inner `Border.settings-card`, which `Styles.Common.axaml:849` insets with `Margin="40,0"`. So the focus ring is drawn ~40px wider than the visible card on each side, floating in the gutter instead of hugging the card, and for grouped cards it will overlap the neighbours. `SettingsAnchor.cs:46-51` hit exactly this and walks down to the inner border for its highlight; the focus visual likely needs the same (focus the inner `Border`, or move the 40px inset onto the outer control).

src/UniGetUI.Avalonia/Views/Controls/Settings/SettingsCard.cs
GabrielDuf: Two things on this breakpoint:

1. **No hysteresis.** The wrap state flips on a single threshold based on the card's own width. Settings pages host these cards in a `StackPanel` inside a `ScrollViewer` with auto scrollbars; wrapping makes every card on the page taller at once, which can bring the vertical scrollbar in, which shrinks available width by ~12px. If the pane width sits in that band around the threshold, you get unwrap → page shortens → scrollbar goes → width grows → wrap again, i.e. a self-sustaining layout loop. A small dead band between the wrap and unwrap thresholds would avoid it.
2. **The measured width isn't the content width.** This measures the outer `SettingsCard`, but `Border.settings-card` carries `Margin="40,0"`, so the effective content breakpoint is ~396, not the CommunityToolkit 476 the constant cites.

src/UniGetUI.Avalonia/Assets/Styles/Styles.Common.axaml
GabrielDuf: Dark-theme `SettingsCardBorderBrush` becomes `Transparent` here (and at `Styles.WindowsMica.axaml:65`), while the light theme keeps `#0F000000`.

Grouped settings cards rely on that brush for their shared divider — e.g. `General.axaml:36-48` and `Administrator.axaml:106-113` stack cards with `CornerRadius="8,8,0,0"` + `BorderThickness="1,0,1,1"`. In dark mode every one of those dividers now paints transparent, so a group of stacked cards renders as one undifferentiated `#0DFFFFFF` block with no separators — asymmetric with light mode, and the opposite of the accessibility goal of the PR.

src/UniGetUI.Avalonia/Assets/Styles/Styles.WindowsMica.axaml
GabrielDuf: `AppOperationsPanelBackground` is set to `Transparent` in the light Mica dictionary, but the dark Mica dictionary (line 70) still has `#66202020`. With the new self-contained operation cards that gives a bare Mica backdrop behind the operations panel in light mode and a tinted panel in dark mode — one of the two was probably meant to change with the other.

src/UniGetUI.Avalonia/Infrastructure/AvaloniaOperationRegistry.cs
GabrielDuf: Keeping the canceled card visible until the user closes it is clearly intentional, but dropping `RemoveAfterDelayAsync` from this path also drops the only reachable call to `IpcOperationApi.ForgetTracking(op.Metadata.Identifier)` — it was only invoked via `Remove` / `RemoveAfterDelayAsync`. A user who cancels operations and never clicks Close now leaves every canceled operation tracked in `IpcOperationApi` and present in `Operations` / `OperationViewModels` for the process lifetime.

Minor follow-on: with the delayed removal gone, the `Interlocked.Exchange(ref cancelHandled, …)` guard below protects only `RemoveProgress`, which is already idempotent, so the guard no longer buys anything.

src/UniGetUI.Avalonia/Assets/Styles/Styles.Common.axaml
```diff
@@ -387,6 +421,26 @@
...
+            </Transitions>
+        </Setter>
+    </Style>
+    <Style Selector="Border.settings-card">
+        <Setter Property="Transitions">
+            <Transitions>
+                <BrushTransition Property="Background" Duration="0:0:0.083"/>
+            </Transitions>
+        </Setter>
+    </Style>
```
GabrielDuf: This transition is the cause of a visible hover regression on every non-Mica platform (reproduced on macOS dark; Windows is unaffected because Mica is active there).

`BrushTransition` interpolates ARGB linearly, and here the two endpoints are not in the same alpha family. Dark theme, `Styles.Common.axaml:319-321`, over the `#202020` page (`:350`):

| | brush | composited |
|---|---|---|
| resting | `#0DFFFFFF` (a=13) | `#2B2B2B` (43) |
| mid-transition (t=0.5) | `#86989898` (a=134) | `#5F5F5F` (95) |
| hover settled | `#323232` (a=255) | `#323232` (50) |

The ramp spikes to 95 and settles at 50 - only 7/255 above resting. In practice the card flashes bright while the pointer is entering or leaving and looks completely un-hovered while the pointer is inside it.

Light theme doesn't spike (both endpoints are near-opaque white) but its settled delta is -5/255, so hover is nearly invisible there too.

Windows is clean because `Styles.WindowsMica.axaml` overrides both brushes in the *same* white channel and varies only alpha - dark `#0DFFFFFF` -> `#15FFFFFF` (`:64`, `:68`), light `#B3FFFFFF` -> `#99FFFFFF` (`:19`, `:23`) - which cannot overshoot. macOS/Linux never merge that dictionary, so they always get the untuned `Styles.Common` pair.

Two fixes needed:
1. Make the non-Mica hover brushes alpha-only overlays like the Mica ones (e.g. dark `SettingsCardHoverBackground` -> `#26FFFFFF`, composites to `#414141`, a visible +22 step), or drop this transition to restore `main`'s instant swap.
2. Separately, the settled deltas in `Styles.Common` (7/255 dark, 5/255 light) are too small to see at all. The Mica dictionary was tuned; `Styles.Common` never was, so every non-Windows platform gets an effectively invisible hover - compounded by `SettingsCardBorderBrush` going `Transparent` in dark, so there is no edge either.

src/UniGetUI.Avalonia/Views/MainWindow.axaml
GabrielDuf: The title now sits in an `Auto` column but keeps `TextTrimming="CharacterEllipsis"`. An `Auto` column is sized to the text's full desired width, so the trimming can never engage.

With a long title — e.g. `KillProcessOperation`'s `"Closing process(es) {0}"` with several process names, or a long installer name — the `*` live-line column collapses to 0 and the grid's desired width exceeds the panel. The hosting `ScrollViewer` has `HorizontalScrollBarVisibility="Disabled"`, so the overflow is clipped rather than scrollable: the Cancel/Close button (col 4) and the `…` button (col 5) get pushed off-screen and become unreachable. Window `MinWidth` is 700, so this is hit at normal sizes.

Previously the title lived in a `*` column, where the ellipsis worked. Suggest giving the title the `*` column (and letting the live line take `Auto`/a capped width), or putting a `MaxWidth` on the title.

src/UniGetUI.Avalonia/Views/Controls/Settings/SettingsCard.cs
GabrielDuf: Worth double-checking this one: `AdornerLayer.DefaultFocusAdorner` *is* set app-wide at `Styles.Common.axaml:406-418`, so a keyboard-focused card does still get a ring even with `settings-card-keyboard-focused` gone. The problem is *where* it lands.

The adorner adorns **this control** — the `SettingsCard` wrapper — not the inner `Border.settings-card`, which `Styles.Common.axaml:849` insets with `Margin="40,0"`. So the focus ring is drawn ~40px wider than the visible card on each side, floating in the gutter instead of hugging the card, and for grouped cards it will overlap the neighbours. `SettingsAnchor.cs:46-51` hit exactly this and walks down to the inner border for its highlight; the focus visual likely needs the same (focus the inner `Border`, or move the 40px inset onto the outer control).

src/UniGetUI.Avalonia/Views/Controls/Settings/SettingsCard.cs
GabrielDuf: Two things on this breakpoint:

1. **No hysteresis.** The wrap state flips on a single threshold based on the card's own width. Settings pages host these cards in a `StackPanel` inside a `ScrollViewer` with auto scrollbars; wrapping makes every card on the page taller at once, which can bring the vertical scrollbar in, which shrinks available width by ~12px. If the pane width sits in that band around the threshold, you get unwrap → page shortens → scrollbar goes → width grows → wrap again, i.e. a self-sustaining layout loop. A small dead band between the wrap and unwrap thresholds would avoid it.
2. **The measured width isn't the content width.** This measures the outer `SettingsCard`, but `Border.settings-card` carries `Margin="40,0"`, so the effective content breakpoint is ~396, not the CommunityToolkit 476 the constant cites.

src/UniGetUI.Avalonia/Assets/Styles/Styles.Common.axaml
GabrielDuf: Dark-theme `SettingsCardBorderBrush` becomes `Transparent` here (and at `Styles.WindowsMica.axaml:65`), while the light theme keeps `#0F000000`.

Grouped settings cards rely on that brush for their shared divider — e.g. `General.axaml:36-48` and `Administrator.axaml:106-113` stack cards with `CornerRadius="8,8,0,0"` + `BorderThickness="1,0,1,1"`. In dark mode every one of those dividers now paints transparent, so a group of stacked cards renders as one undifferentiated `#0DFFFFFF` block with no separators — asymmetric with light mode, and the opposite of the accessibility goal of the PR.

src/UniGetUI.Avalonia/Assets/Styles/Styles.WindowsMica.axaml
GabrielDuf, line 25: `AppOperationsPanelBackground` is set to `Transparent` in the light Mica dictionary, but the dark Mica dictionary (line 70) still has `#66202020`. With the new self-contained operation cards that gives a bare Mica backdrop behind the operations panel in light mode and a tinted panel in dark mode — one of the two was probably meant to change with the other.

src/UniGetUI.Avalonia/Infrastructure/AvaloniaOperationRegistry.cs
GabrielDuf: Keeping the canceled card visible until the user closes it is clearly intentional, but dropping `RemoveAfterDelayAsync` from this path also drops the only reachable call to `IpcOperationApi.ForgetTracking(op.Metadata.Identifier)` — it was only invoked via `Remove` / `RemoveAfterDelayAsync`. A user who cancels operations and never clicks Close now leaves every canceled operation tracked in `IpcOperationApi` and present in `Operations` / `OperationViewModels` for the process lifetime.

Minor follow-on: with the delayed removal gone, the `Interlocked.Exchange(ref cancelHandled, …)` guard below protects only `RemoveProgress`, which is already idempotent, so the guard no longer buys anything.

src/UniGetUI.Avalonia/Assets/Styles/Styles.Common.axaml
GabrielDuf, line 459: This transition is the cause of a visible hover regression on every non-Mica platform (reproduced on macOS dark; Windows is unaffected because Mica is active there).

`BrushTransition` interpolates ARGB linearly, and here the two endpoints are not in the same alpha family. Dark theme, `Styles.Common.axaml:319-321`, over the `#202020` page (`:350`):

| | brush | composited |
|---|---|---|
| resting | `#0DFFFFFF` (a=13) | `#2B2B2B` (43) |
| mid-transition (t=0.5) | `#86989898` (a=134) | `#5F5F5F` (95) |
| hover settled | `#323232` (a=255) | `#323232` (50) |

The ramp spikes to 95 and settles at 50 - only 7/255 above resting. In practice the card flashes bright while the pointer is entering or leaving and looks completely un-hovered while the pointer is inside it.

Light theme doesn't spike (both endpoints are near-opaque white) but its settled delta is -5/255, so hover is nearly invisible there too.

Windows is clean because `Styles.WindowsMica.axaml` overrides both brushes in the *same* white channel and varies only alpha - dark `#0DFFFFFF` -> `#15FFFFFF` (`:64`, `:68`), light `#B3FFFFFF` -> `#99FFFFFF` (`:19`, `:23`) - which cannot overshoot. macOS/Linux never merge that dictionary, so they always get the untuned `Styles.Common` pair.

Two fixes needed:
1. Make the non-Mica hover brushes alpha-only overlays like the Mica ones (e.g. dark `SettingsCardHoverBackground` -> `#26FFFFFF`, composites to `#414141`, a visible +22 step), or drop this transition to restore `main`'s instant swap.
2. Separately, the settled deltas in `Styles.Common` (7/255 dark, 5/255 light) are too small to see at all. The Mica dictionary was tuned; `Styles.Common` never was, so every non-Windows platform gets an effectively invisible hover - compounded by `SettingsCardBorderBrush` going `Transparent` in dark, so there is no edge either.

GabrielDuf (changes requested):
I left a few comments. These changes break the settings-card hover on every non-Mica configuration — macOS, Linux, Windows 10, and Windows 11 with transparency effects turned off. Windows 11 with transparency on is the only setup that's safe, because Styles.WindowsMica overrides the two brushes involved; Styles.Common never got the same tuning. You can reproduce it on your own machine: turn off Personalization → Colors → Transparency effects, restart, and hover a settings card in dark mode.

GabrielDuf:
Here's a video showing the issue I mentioned.

Also, I think the hover color is now way too light. It looks a bit weird.

https://github.com/user-attachments/assets/0849b463-69c1-4bfd-aed5-fc68be5405ee

Meteoni:
That isn't supposed to happen. I'll look into it. 

https://github.com/user-attachments/assets/f8523d9b-2189-4cf5-805f-3abca9b9ca0c

Meteoni:
@GabrielDuf I've updated the PR with some additional fixes bundled in for several pre-existing bugs that I found along the way. Appreciate your review.

